PR #7, opened 2020-02-25
Improved RadioGen

plasorak:
The changes in the module allow it to:
- Does time-correlated decays (Bismuth/Polonium),
- Uses a `struct` to pass the decay info, rather than `std::tuple`,
- Better parsing of file spectrum files (convert the `TGraph` to the `TH1D` without relying on the format of the file),
- Creates output plots (time of the particules generated, position (xy, yz), time correlation, momentum, pdg),
- Can be called with volume names, rather than hardcoding the box values based on geometry.

Probably would need to be tested more systematically to make sure that it yields to the same results as the previous version. I can do make some plots that compare what is generated for different experiments if required (I'd just need to be told which actual fhicl are to be run).

### DIFF
--- a/larsim/EventGenerator/RadioGen_module.cc
+++ b/larsim/EventGenerator/RadioGen_module.cc
@@ -112,14 +112,14 @@ namespace evgen {
     void readfile(std::string type, std::string const& filename);
     void samplespectrum(std::string nuclideName, ParticleInfo& part);
 
-    
+
     struct ParticleInfo{
       ti_PDGID pdg;
       td_Mass mass;
       TLorentzVector pos;
       TLorentzVector mom;
     };
-    
+
     ParticleInfo AlphaDecay(double t, TVector3 pos, double time);
     ParticleInfo BetaDecay(double t, TVector3 pos, double time, double Z);
     std::vector<ParticleInfo> BetaThenAlphaDecay(double t_beta, double t_alpha, double t_mean, TVector3 pos, double time, double Z);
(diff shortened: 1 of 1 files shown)


plasorak:
```
srcs/larsim $ removeWhiteSpace.sh < changed_files
bash: removeWhiteSpace.sh: command not found...
```

lgarren:
@plasorak 
You need to setup a recent larsoft.  larsoft v08_41_01 or later will have removeWhiteSpace.sh

lgarren:
Also, those instructions are not quite right.  Cat changed_files and then 
removeWhiteSpace.sh <list of files>

lgarren:
You have to give removeWhiteSpace.sh the list of files by hand.

plasorak:
Thanks, done.

knoepfel:
Hi @plasorak, thanks for the PR.  Looks like there are some improvements that can yet be made.  One of the issues, though, is that this code required some improvement before you started working on it.  So, I can either make suggestions that are mostly specific to your contributions, or we can take a broader approach and try to clean up the overall design.  What is your opinion?

tomjunk:
Is this PR moving along?

knoepfel:
@tomjunk, this is one of several PRs that needs some further development.  Pierre's collaborators in Sussex have been contacted regarding this.  We are awaiting word from them whether they are ready to proceed in addressing any issues brought up here.

plasorak:
Hi @knoepfel, I think we should ask Juergen Reichenbacher (juergen) or Jason Stock (jstock) for what to do, since they are in charge of this module. I agree with your comment. I'll send them an email and CC you.

plasorak:
Hi @knoepfel, since I'm not getting any answer, let's go ahead and make the changes that you are suggesting.
Since I've just learnt how to do it, I think it'd be good to implement unit tests as well. I'll start with this.

knoepfel:
@plasorak, starting with unit tests is an excellent plan.  I'll look through the code more in the next day or so and give you some of my own thoughts regarding it.

plasorak:
Actually, we discussed at the DAQ Physics Performance meeting yesterday that we may use an external library to do this, RadioGen may not be generic enough to do all the decays we need. So probably better to put this on hold. Sorry for that!

knoepfel:
No problem.  Would you like to keep this PR "under discussion" or would it be best to just close it?

plasorak:
I think it's probably better to keep this under discussion. :)

knoepfel:
Okay, will do.

lgarren:
This pull request is superseded by larsim PR #23 
Closing

PR #10, opened 2020-03-17
Small fix to interpretation of azimuthal angle in MUSUN generator.

vpec0:
The input flux assumes convention where azimuth angle (phi) runs counterclockwise from east direction. This translates to counterclockwise rotation from local negative Z direction to local negative X. The code as it is now makes phi run clockwise. To add here, flux angles determine direction of the origin of the muons, the direction of their momentum is the opposite. Hence direction cosines cx and cz in the code should both have a plus sign.

### DIFF
--- a/larsim/EventGenerator/MuonPropagation/MUSUN_module.cc
+++ b/larsim/EventGenerator/MuonPropagation/MUSUN_module.cc
@@ -549,7 +549,7 @@ namespace evgen{
 
     //  The minus sign above is for y-axis pointing up, so the y-momentum
     //  is always pointing down
-    cx       = -sin(theta)*sin(phi);
+    cx       = +sin(theta)*sin(phi);
     cy       = -cos(theta);
     cz       = +sin(theta)*cos(phi);
     Momentum = std::sqrt(Energy*Energy-m*m); // Get momentum


lgarren:
trigger build

tomjunk (commented):
This looks like a good change, but I think it should be reviewed by Karl Warburton and Vitaly Kudryavtsev.  LBNE-Doc-9673-v1 shows on p. 14 a comparison of the phi angle and the local surface topography at SURF.  From that note, theta looks like the zenith angle and phi the azimuth angle, of a ray starting at the detector pointing back at the muon's origin, so opposite to the momentum as Viktor says.  I will forward this to Vitaly and Karl.

lgarren:
trigger build

tomjunk (approved):
Both Vitaly and Karl have reviewed the change and approve.  It was discussed at a DUNE calibration meeting and found necessary.

lgarren:
thanks Tom.

tomjunk:
Approve

lgarren:
trigger build

lgarren:
dunetpc may need to make a change to accommodate this fix.  However, since they have signed off, I will approve and merge.

lgarren:
approve

PR #12, opened 2020-04-06
Fixing a bug in MUSUN generator

vpec0:
Description of the bug: 
MUSUN generates muons on the surface of a box which contains the main detector module and part of the surrounding rock. 

For muons pointing in the positive direction of Z, the direction of FNAL beam, the face at Max Z is chosen. For muons pointing in the negative direction of Z, the direction towards FNAL, the face at Min Z is chosen. This effectively makes the muons be shot away from the main volume on these faces.

This is not the case for the XY faces with direction along X.

The author of the code, Karl Warburton, has been contacted and the fix was agreed on.

This bug is not directly related to the previous fix in MUSUN, 8f560d4b29eb.


lgarren:
trigger build

yangtj207:
+1

knoepfel:
approve

PR #15, opened 2020-04-30
New module `POTaccumulator` counting the total POT in a simulated dataset.

PetrilloAtWork:


tomjunk:
Is it just me, or does the EXCLUDE line in the CMakeLists.txt file

https://raw.githubusercontent.com/PetrilloAtWork/larsim/feature/gp_POTaggregator/larsim/EventGenerator/CMakeLists.txt

art_make(
         EXCLUDE
           "POTaccumulator_module.cc"
...

prevent compilation of your new module?

PetrilloAtWork:
It prevents `art_make()` from scheduling it for building. But note on line 34,
`simple_plugin(POTaccumulator "module"
  larcoreobj_SummaryData
  ${MF_MESSAGELOGGER}
  )`
which pulls the module back in.
I choose this way to have a lighter linkage for the new module, since it requires quite fewer libraries than the siblings.

tomjunk:
Thanks!  I should read carefully and learn tricks.  

Is there a reason for the DumpMCParticles reference in the fcl file?

PetrilloAtWork:
Oh, yes, absolutely. Very strong reason indeed.
I started with `dump_mcparticles.fcl` as template, and I missed to change that configuration key.
I hope that explanation on why `DumpMCParticles` is absolutely necessary there satisfies you.

And also: fixed.
See? you are already reading it more carefully than I had.

tomjunk:
Looks fine to me, though I cannot run the program in my head.  To say a a number of subruns "miss POT data" is a bit odd.  "are missing POT data" or "lack POT data" is better.

tomjunk:
approved

larsim/EventGenerator/CMakeLists.txt
```diff
@@ -1,4 +1,7 @@
-art_make(LIB_LIBRARIES
+art_make(
+         EXCLUDE
+           "POTaccumulator_module.cc"
```
knoepfel, line 3: Quotes not needed.

larsim/EventGenerator/POTsummary.fcl
```diff
@@ -0,0 +1,113 @@
...
+      SummaryCategory:    "POTsummary"
+      RunSummaryCategory: "POTsummary"
+
+    } # potsummary
+  } # analyzers
+
+  summaries: [ potsummary ]
+
+  trigger_paths: []
+  end_paths:     [ summaries ]
```
knoepfel: These two lines are unnecessary.

larsim/EventGenerator/POTaccumulator_module.cc
```diff
@@ -0,0 +1,307 @@
...
+} // sim::POTaccumulator::printMissingSubrunList()
+
+
+//------------------------------------------------------------------------------
+void sim::POTaccumulator::printRunSummary() const {
+
+  // count subruns in run
+  std::map<art::RunID, unsigned int> subrunCount;
+  for (art::SubRunID const& ID: util::get_elements<0U>(fPresentSubrunFragments))
+    ++subrunCount[art::RunID{ ID.run() }];
```
knoepfel: Or just `++subrunCount[ID.runID()]`.

larsim/EventGenerator/POTaccumulator_module.cc
```diff
@@ -0,0 +1,307 @@
...
+
+} // sim::POTaccumulator::printMissingSubrunList()
+
+
+//------------------------------------------------------------------------------
+void sim::POTaccumulator::printRunSummary() const {
+
+  // count subruns in run
+  std::map<art::RunID, unsigned int> subrunCount;
+  for (art::SubRunID const& ID: util::get_elements<0U>(fPresentSubrunFragments))
```
knoepfel: I encourage:

```.cpp
#include "ranges/v3/view.hpp"
...
for (auto::SubRunID const& ID : fPresentSubrunFragments | ranges::view::keys) {
  ...
}
```
This is the direction the standard is headed, and changing to the STL should be straightforward once C++20 is supported.

larsim/EventGenerator/POTaccumulator_module.cc
```diff
@@ -0,0 +1,307 @@
...
+  POTaccumulator(POTaccumulator &&) = delete;
+  POTaccumulator& operator = (POTaccumulator const&) = delete;
+  POTaccumulator& operator = (POTaccumulator &&) = delete;
+
+
+  // Nothing to be done at event level.
+  virtual void analyze(art::Event const& event) override {}
+
+  /// Collects information from each subrun.
+  virtual void beginSubRun(art::SubRun const& subRun) override;
```
knoepfel: By choosing `beginSubRun` instead of `endSubRun`, this module can never be used in a job that creates `POTsummary` objects.  Is there a reason why the functionality of `beginSubRun` cannot be moved to `endSubRun`?

larsim/EventGenerator/POTaccumulator_module.cc
```diff
@@ -0,0 +1,307 @@
...
+
+    printRunSummary();
+
+  } // if run report requested
+
+
+  //
+  // print the total summary
+  //
+  printSummary();
```
knoepfel, line 229: The function name nearly says exactly what the comments say.  If you change the function name to `printTotalSummary()`, then you can be rid of the comments.  Comments have to be maintained, too.

larsim/EventGenerator/POTaccumulator_module.cc
```diff
@@ -0,0 +1,307 @@
...
+  ++fPresentSubrunFragments[ID];
+
+  //
+  // accumulate the information by run
+  //
+  auto const& runID = subRun.getRun().id();
+
+  sumdata::POTSummary const& subRunPOT = *summaryHandle;
+
+  fRunPOT[runID].aggregate(subRunPOT);
```
knoepfel: These lines can be collapsed to`fRunPOT[subRun.runID()].aggregate(*summaryHandle)`.  You should also consider protecting this function with the `art::disjoint_ranges(...)` API if you are going to handle product-aggregation yourself.

larsim/EventGenerator/POTaccumulator_module.cc
```diff
@@ -0,0 +1,307 @@
...
+  log << size(fMissingSubrunFragments) << " subruns miss POT information:";
+
+  auto const efound = fPresentSubrunFragments.cend();
+
+  for (auto const& [ id, nMissing ]: fMissingSubrunFragments) {
+
+    // add to the count of fragments the ones which did have
+    unsigned int nFragments = nMissing;
+    auto const iFound = fPresentSubrunFragments.find(id);
+    if (iFound != efound) nFragments += iFound->second;
```
knoepfel: This logic doesn't make sense to me.  Under what circumstance do you expect a `SubRunID` that does *and* does not contain a POT summary object?  And if such a situation can occur, why report the total number of fragments instead of just the number that are actually missing?

larsim/EventGenerator/POTaccumulator_module.cc
```diff
@@ -0,0 +1,307 @@
...
+  // print the run summary
+  //
+
+  if (!fRunOutputCategory.empty()) {
+
+    if (!fMissingSubrunFragments.empty()) printMissingSubrunList();
+
+    printRunSummary();
+
+  } // if run report requested
```
knoepfel: The comment on this line seems unnecessary.

larsim/EventGenerator/POTaccumulator_module.cc
```diff
@@ -0,0 +1,307 @@
...
+ * @brief  Prints on console the total Protons On Target from the input subruns.
+ *
+ * This module collects information from each of the subrun in the input files
+ * and prints the total of the protons on target. It also optionally prints the
+ * total for each observed run.
+ * If a subrun is met more than once, the information from all subrun fragments
+ * are added together (i.e. it is assumed that summary information is
+ * complementary rather than duplicate).
+ *
+ * The output is printed to the console via message facility.
```
knoepfel: As you know, the message facility can be configured so that nothing is printed to the console.  Should adjust to say "The output can be printed to the console via the appropriate message facility configuration."

larsim/EventGenerator/POTaccumulator_module.cc
```diff
@@ -0,0 +1,307 @@
...
+
+/**
+ * @brief  Prints on console the total Protons On Target from the input subruns.
+ *
+ * This module collects information from each of the subrun in the input files
+ * and prints the total of the protons on target. It also optionally prints the
+ * total for each observed run.
+ * If a subrun is met more than once, the information from all subrun fragments
+ * are added together (i.e. it is assumed that summary information is
+ * complementary rather than duplicate).
```
knoepfel, line 38: How familiar are you with the framework's automatic product aggregation handling for (sub)runs?

https://cdcvs.fnal.gov/redmine/projects/art/wiki/Run_and_SubRun_products

The framework does careful bookkeeping to ensure that a (sub)run product presented to the user represents either an entire (sub)run or a well-formed range of validity.  Unless users are inserting `POTsummary` products in their files with different semantics (i.e. `full(Sub)Run` vs. `(sub)RunFragment`), the `POTsummary` objects will be complementary/disjoint.

Your code below does not check for that explicitly, even though it could.  Let me know if you're interested in how to do it.

larsim/EventGenerator/POTsummary.fcl
```diff
@@ -0,0 +1,113 @@
...
+      } # LogStandardOut
+
+    } # destinations
+  } # message
+} # services
+
+
+physics: {
+  producers:{}
+  filters:  {}
```
knoepfel: Unnecessary lines

knoepfel (commented):
This module seems sensible considering that the framework does not support subrun-product aggregation across subrun boundaries.  At some point, it may become useful to make a feature request of the framework.

Some comments are admittedly nitpicky.  See what you think, though.

larsim/EventGenerator/POTsummary.fcl
PetrilloAtWork: Since when? _art_ news, or I was always mistaken?

knoepfel: The `trigger_paths` and `end_paths` parameters are only necessary when you need to *exclude* paths from execution.  In other words, if the `trigger_paths` and `end_paths` parameters are specified, then only the paths listed in the sequence will be executed.  If you intend all paths to be executed, then you can omit those parameters and *art* will figure it out.

When I joined the art team in 2014, this was the behavior...and it hasn't changed.

larsim/EventGenerator/CMakeLists.txt
PetrilloAtWork: ... but I prefer to use them :stuck_out_tongue:
Also, learning GitHub here. Please forgive.

larsim/EventGenerator/POTsummary.fcl
PetrilloAtWork: ~What about `filters: {}` and `producers: {}`? is it recommended to leave them there?~
Found the answer in a later coment.

knoepfel: As I say in a later comment, there is no need to have the empty `filters` and `producers` tables.

larsim/EventGenerator/POTaccumulator_module.cc
PetrilloAtWork: In case one input file with a fragment of that subrun (from one input file) contains such information, and another (from a different input file) does not.
I think there is some value in seeing at once that a subrun is _partially_ missing information.
To be honest, I think that may be an error where files from different samples are mixed, with overlapping subrun numbers.
This all section of missing information is a debugging tool: IMO this information should not be missing. But I don't feel that strong enough to throw exceptions.

larsim/EventGenerator/POTsummary.fcl
PetrilloAtWork: Implemented.

larsim/EventGenerator/POTaccumulator_module.cc
PetrilloAtWork: As long as it works without me having to go through hundreds lines of compiler error message, I am happy with this: implemented.
It's `range/v3/view.hpp`, though.

larsim/EventGenerator/POTaccumulator_module.cc
PetrilloAtWork: Implemented.

larsim/EventGenerator/POTaccumulator_module.cc
PetrilloAtWork: Excellent point. There was no specific reason beyond flipping a coin. Implemented.

larsim/EventGenerator/POTaccumulator_module.cc
PetrilloAtWork: Reduced. I use `// if` as a graphical reminder of the block.

larsim/EventGenerator/POTaccumulator_module.cc
PetrilloAtWork: Implemented.

larsim/EventGenerator/POTsummary.fcl
PetrilloAtWork: Implemented.

larsim/EventGenerator/POTaccumulator_module.cc
PetrilloAtWork: Ok.

larsim/EventGenerator/POTaccumulator_module.cc
PetrilloAtWork: Going from a `art::SubRun` to a `art::RunID` seems to be a two step process, either via `art::Run` (`art::SubRun::run()`) or via `art::SubRunID` (`art::SubRun::id()`).
The direct one is a `private` road :no_entry: do not trespass.

----

Back to us: I have reimplemented the aggregation with `art::SummedData`, which appears to be proper. The implementation in this module ends up "surpassing" _art_'s in the sense that it checks overlaps also across files, while _art_ deliberately does not.

Unfortunately `art::SummedData` interface falls short of my need: the original module design would aggregate all runs at the end of the job, but that is not possible with `art::SummedData` which `update()`s only via `art::ValidHandle`s. An interface with `art::SummedData::update(T const&, RangeSet const&)` or `art::SummedData::update(art::SummedData<T> const&)` would have helped.
I close to redesign that feature, gaining the protection against duplicates but at the cost of doing the aggregation twice.

I gave the module two input files with the same events. It did not blink, and it performed no double count. :sunglasses:

ghost:
trigger build

lgarren:
Reporting here that the CI tests are normal.

larsim/EventGenerator/POTaccumulator_module.cc
knoepfel: You still left the comment in there. 😉

ghost:
approve

lgarren:
approve

PR #16, opened 2020-05-04
SingleGen module now adds `RunData` information as per LArSoft protocol.

PetrilloAtWork:
`SingleGen` does currently not write `sumdata::RunData` data product, which foils the already loosy geometry version check.
This change makes `SingleGen` create that data product just like `GENIEGen` does (as in: I copied the code from there).
This is not a breaking change; only new samples will be affected.

### DIFF
--- a/larsim/EventGenerator/SingleGen_module.cc
+++ b/larsim/EventGenerator/SingleGen_module.cc
@@ -508,8 +508,9 @@ namespace evgen{
   //____________________________________________________________________________
   void SingleGen::beginRun(art::Run& run)
   {
-    art::ServiceHandle<geo::Geometry const> geo;
-    run.put(std::make_unique<sumdata::RunData>(geo->DetectorName()));
+    geo::GeometryCore const& geom = *(lar::providerFrom<geo::Geometry>());
+    run.put
+      (std::make_unique<sumdata::RunData>(geom.DetectorName()), art::fullRun());
   }
 
 
--- a/larsim/EventGenerator/SingleGen_module.cc
+++ b/larsim/EventGenerator/SingleGen_module.cc
@@ -40,6 +40,12 @@
 #include "nusimdata/SimulationBase/MCTruth.h"
 #include "nusimdata/SimulationBase/MCParticle.h"
 
+// LArSoft libraries
+#include "larcore/Geometry/Geometry.h"
+#include "larcorealg/Geometry/GeometryCore.h"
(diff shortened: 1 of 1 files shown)


lgarren:
trigger build

larsim/EventGenerator/SingleGen_module.cc
```diff
@@ -490,6 +500,16 @@ namespace evgen{
...
+
+  }
+
+
+  //____________________________________________________________________________
+  void SingleGen::beginRun(art::Run& run)
+  {
+    art::ServiceHandle<geo::Geometry const> geo;
+    run.put(std::make_unique<sumdata::RunData>(geo->DetectorName()));
   }
```
knoepfel: I suggest that if this product is intended to represent an entire run, then you be explicit about the range-of-validity:

```.cpp
run.put(std::make_unique<RunData>(...), art::fullRun());
```

It is then quite explicit what is going on.  With a future version of *art*, I hope to make the second argument required.

PetrilloAtWork: In my interpretation it is not clear whether `RunData` is conceptually intended to be full run or run "fragment", but the _current_ content is definitely run-wide. In fact, it should be job-wide, but `art::Results` are cast out of standard modules.
I have implemented the suggestion.

absolution1:
approve

lgarren:
approve

yangtj207:
trigger build

ghost:
The clang CI builds are failing with an error in SingleGen_module, see: [https://dbweb5.fnal.gov:8443/LarCI/app/storage/docs/2020/05/06/buildLArSoft%238eGHZg2.log](url)

> 1832: /scratch/workspace/lar_ci/label_exp/SLF6/label_exp2/swarm/LArSoft/srcs/larsim/larsim/EventGenerator/SingleGen_module.cc:229:10: error: 'produce' overrides a member function but is not marked 'override' [-Werror,-Winconsistent-missing-override]
1833:     void produce(art::Event& evt);
1834:          ^
1835: /cvmfs/larsoft.opensciencegrid.org/products/art/v3_05_00/include/art/Framework/Core/EDProducer.h:67:18: note: overridden virtual function is here
1836:     virtual void produce(Event&) = 0;
1837:                  ^
1838: 1 error generated.

lgarren:
The problem was fixed after this PR was merged with develop.  The head of develop should be fine at this point.

PR #18, opened 2020-05-19
Added new Correlated ion/scint calculation alg for anticorrelation between electrons and photons

wforeman:


lgarren:
trigger build

larsim/LegacyLArG4/ISCalculationCorrelated.cxx
```diff
@@ -0,0 +1,153 @@
...
+
+  //----------------------------------------------------------------------------
+  ISCalculationCorrelated::ISCalculationCorrelated(CLHEP::HepRandomEngine&)
+  {
+  }
+
+  //----------------------------------------------------------------------------
+  ISCalculationCorrelated::~ISCalculationCorrelated()
+  {
+  }
```
knoepfel: Suggest removing destructor definition as no work is being done.

larsim/IonizationScintillation/IonAndScint_module.cc
```diff
@@ -107,6 +109,10 @@ namespace larg4
         {
             fISAlg = new ISCalcSeparate();
         }
+        else if(calcTag.label().compare("Correlated") == 0)
+        {
+            fISAlg = new ISCalcCorrelated();
+        }
```
knoepfel: This will work just fine and do what is required.  Are you willing to polish some of this design (which was put in place before your contributions here)?  If so, I would recommend something like:
```.cpp
if (calcTag.label() == "NEST") {
  fISAlg = std::make_unique<ISCalcNESTLAr>(fEngine);
}
else if (calcTag.label() == "Separate") {
  ...
}
...
```
With this approach, you can directly compare the `label()` with the string literal using `==`.  Also, the `fISAlg` object would be of type `std::unique_ptr<ISCalc>`, which would mean that the `delete fISAlg` statement later on would be unnecessary as that would happen automatically whenever the class is destroyed.

Lastly, I argue all of the `beginJob` functionality can be moved to the constructor.  No reason to have to implement a separate function to initialize the class.

I know these design issues are not particularly related to your PR.  But while you were in this "neck of the woods," I thought you might be willing to consider it.

larsim/IonizationScintillation/IonAndScint_module.cc
```diff
@@ -123,10 +129,7 @@ namespace larg4
     {
         std::cout << "IonAndScint endJob." << std::endl;
 
-        if(fISAlg)
-        {
-            delete fISAlg;
-        }
+        if(fISAlg) delete fISAlg;
```
knoepfel, line 119: See comment above about using `std::unique_ptr<ISCalc>` instead as the type of `fISAlg`.

larsim/IonizationScintillation/ISCalcCorrelated.cxx
```diff
@@ -0,0 +1,146 @@
...
+
+        if (fSCE->EnableSimEfieldSCE())
+        {
+            eFieldOffsets = fSCE->GetEfieldOffsets(pos);
+            EField = std::sqrt((efield + efield*eFieldOffsets.X())*(efield + efield*eFieldOffsets.X())
+                              +(efield*eFieldOffsets.Y()*efield*eFieldOffsets.Y())
+                              +(efield*eFieldOffsets.Z()*efield*eFieldOffsets.Z()) );
+        }
+
+        return EField;
```
knoepfel: Suggest writing this more concisely as:
```.cpp
double ISCalc::Correlated::EfieldAtStep(...) const
{
  if (not fSCE->EnabelSimEfieldSCE()) {
    return efield;
  }

  auto const eFieldOffsets = fSce->GetEfieldOffsets(edep.MidPoint());
  return efield * std::hypot(1 + eFieldOffsets.X(), eFieldOffsets.Y(), eFieldOffsets.Z());
}
```

larsim/LegacyLArG4/ISCalculationCorrelated.cxx
```diff
@@ -0,0 +1,153 @@
...
+    return;
+  }
+
+  //----------------------------------------------------------------------------
+  // fNumIonElectrons returns a value that is not corrected for life time effects
+  void ISCalculationCorrelated::Reset()
+  {
+    fEnergyDeposit   = 0.;
+    fNumScintPhotons = 0.;
+    fNumIonElectrons = 0.;
```
knoepfel, line 76: The `fEnergyDeposit` and `fNumIonElectrons` data members are only used within the `calculateIonizationAndScintillation` function and should therefore be removed as class data members.  It is much better to define these variables only in the function where they are used.

larsim/LegacyLArG4/ISCalculationCorrelated.cxx
```diff
@@ -0,0 +1,153 @@
...
+  {
+  }
+
+  //----------------------------------------------------------------------------
+  ISCalculationCorrelated::~ISCalculationCorrelated()
+  {
+  }
+
+  //----------------------------------------------------------------------------
+  void ISCalculationCorrelated::Initialize()
```
knoepfel: Suggest placing all initialization in the constructor instead of in a separate `Initialize` function.

larsim/LegacyLArG4/ISCalculationCorrelated.h
```diff
@@ -0,0 +1,51 @@
...
+namespace larg4 {
+
+ class ISCalculationCorrelated : public ISCalculation {
+
+ public:
+
+   ISCalculationCorrelated(CLHEP::HepRandomEngine&);
+   virtual ~ISCalculationCorrelated();
+
+   void   Initialize();
```
knoepfel: See comment about moving initialization to constructor.

larsim/IonizationScintillation/ISCalcCorrelated.cxx
```diff
@@ -0,0 +1,146 @@
...
+{
+    //----------------------------------------------------------------------------
+    ISCalcCorrelated::ISCalcCorrelated()
+    {
+    }
+
+    //----------------------------------------------------------------------------
+    ISCalcCorrelated::~ISCalcCorrelated()
+    {
+    }
```
knoepfel: Suggest removing this definition as no work is being done.

larsim/IonizationScintillation/ISCalcCorrelated.cxx
```diff
@@ -0,0 +1,146 @@
...
+
+    //----------------------------------------------------------------------------
+    void ISCalcCorrelated::Initialize()
```
knoepfel: Suggest moving all initialization to the constructor.

larsim/IonizationScintillation/ISCalcCorrelated.cxx
```diff
@@ -0,0 +1,146 @@
...
+        return;
+    }
+
+    //----------------------------------------------------------------------------
+    void ISCalcCorrelated::Reset()
+    {
+        fEnergyDeposit           = 0.;
+        fNumScintPhotons         = 0.;
+        fNumIonElectrons         = 0.;
+        fScintillationYieldRatio = 0.;
```
knoepfel, line 54: Any variables that are used in only one function should be defined in only that function--e.g. `fEnergyDeposit` need not be a class member variable as only `CalcIonAndScint` uses it.

knoepfel (commented):
The added functionality is useful, thanks.  I have made several comments regarding code implementation.  Please take a look, and I'm happy to discuss with you.

wforeman:
Hi Kyle,

Thanks for the suggestions. Was mostly copying the conventions laid out already in the existing algs, but I’m more than happy to spruce things up while I’m at it. Should I just go ahead and do another pull request when I’m done?

However, I do have concerns with this suggestion:
> In larsim/LegacyLArG4/ISCalculationCorrelated.cxx <https://github.com/LArSoft/larsim/pull/18#discussion_r428022144>:
> 
> > +    fEnergyDeposit   = 0.;
> +    fNumScintPhotons = 0.;
> +    fNumIonElectrons = 0.;
> The fEnergyDeposit and fNumIonElectrons data members are only used within the calculateIonizationAndScintillationfunction and should therefore be removed as class data members. It is much better to define these variables only in the function where they are used.
> 
> 

I’m not sure I agree with you here — the alg classes (ISCalculation[Method]) inherit from the ISCalculation class, which contains these public getter functions:
EnergyDeposit()              const { return fEnergyDeposit;   }
NumberIonizationElectrons()  const { return fNumIonElectrons; }
NumberScintillationPhotons() const { return fNumScintPhotons; }

These functions, which return number of ions and photons for each G4 step, are used in other modules, for example LArVoxelReadout.cxx, where the electron drift is carried out in LegacyLArG4. Similarly, in OpFastScintillation.cxx and IonizationAndScintillation.cxx use NumberScintillationPhotons().  The situation is similar in the refactorized code too. So I think these are meant to be member variables.

lgarren:
@wforeman  please make changes in this pull request.  The process is straightforward, you just commit changes on the existing fork.

wforeman:
Sounds good, thank you Lynn. 
I’ve pushed the changes to the fork, so the PR can continue.

Best,
Will

lgarren:
trigger build

knoepfel:
Hi Will, re. https://github.com/LArSoft/larsim/pull/18#issuecomment-631506083, I didn't see that those member variables were used in the base class.  Feel free to disregard that comment, then.  Although, as an aside, protected data is not usually a good idea.

aihimmel:
I wouldn't stress too much about redesigning the access to number of photons and number of electrons. This was already done in the move to the new LArG4, so no need to try to redesign the behavior here.

Speaking of which, Will, would you consider implementing this in the new LArG4 as well? Most experiments are moving that way, and it seems like it would be much easier for you to make the necessary changes there than to try to find someone else to redo your work.

wforeman:
Hi Alex,

This new “Correlated” class was added to both LegacyLArG4 and the “new LArG4" (i.e., larsim/IonizationAndScintillation). I haven’t been able to test it out in the new LArG4, but it exactly mirrors the existing algs (ISCalcSeparate, ISCalcNESTLAr) so it should work.

Will

aihimmel:
Excellent, thanks Will. Clearly I haven't had a chance to look through the PR carefully. I will try to do so soon (though Kyle probably has a better eye than me ;) ).

larsim/IonizationScintillation/ISCalcCorrelated.cxx
```diff
@@ -0,0 +1,124 @@
...
+
+        return;
+
+    }
+
+
+    //----------------------------------------------------------------------------
+    double ISCalcCorrelated::GetScintYieldRatio(sim::SimEnergyDeposit const& edep)
+    {
+      return fNumScintPhotons / edep.Energy();
```
aihimmel: This is not what is meant by `ScintYieldRatio`. This ratio is supposed to be the fraction of fast/singlet photons in the scintillation. There's some confusing names/comments in SimEnergyDeposit that might be worth fixing, but for now let's at least get the right value here.  This is what's used in the `ISCalcSeparate.cxx` around line 190:
```suggestion
        if (!fLArProp->ScintByParticleType())
        {
            return fLArProp->ScintYieldRatio();
        }
        switch (edep.PdgCode())
        {
            case  2212:
               return fLArProp->ProtonScintYieldRatio();
            case  13:
            case -13:
                return fLArProp->MuonScintYieldRatio();
            case  211:
            case -211:
                return fLArProp->PionScintYieldRatio();
            case  321:
            case -321:
                return fLArProp->KaonScintYieldRatio();
            case 1000020040:
                return fLArProp->AlphaScintYieldRatio();
            case  11:
            case -11:
            case  22:
                return fLArProp->ElectronScintYieldRatio();
            default:
                return fLArProp->ElectronScintYieldRatio();
        }
    }
```

aihimmel (changes requested):
I've looked through everything. There is one comment pointing out a significant problem which will need to be fixed before this PR can be merged. Otherwise, everything else looks fine to me.

lgarren:
@wforeman  Have you had a chance to look at Alex's concerns yet?

wforeman:
Hi Lynn — yes, I’m correcting it now. I’ll have the branch updated soon.

wforeman:
Hi Lynn and others, 

I’d like to merge the develop branch of larsim back into my branch before pushing this latest change, but I’m not sure how to do this now that I’ve forked the repo. Would someone be able to help me with this? I tried “git pull origin develop” but it just references the forked repo in my own github account…

Thanks,
Will

lgarren:
Will, you need "git pull upstream develop"

________________________________________

aihimmel (approved):
Looks good to me. Approve.

wforeman:
Hi all,

The PR is now updated. Thanks to Alex for pointing out that error! Sorry, I was confusing “ScintYieldRatio" with “ScintYield” which refers to photons/MeV.

Speaking of this, while this goes beyond my PR, may I suggest that this terminology be changed? Not only does it look very similar to the unrelated “ScintYield” parameter, but in most literature (for example https://arxiv.org/abs/1511.07721 <https://arxiv.org/abs/1511.07721>) the metric quoted from measurements is the singlet-to-triplet ratio, S/T, whereas in the code ScintYieldRatio is the fraction of singlet to total light, S/(S+T). The most often quoted S/T ratio for MIPs is 0.3, which is what ScintYieldRatio is set to in PhotonPropagation/prodsingle_libraryana.fcl, and possibly more generally in lardata:
https://www.phy.bnl.gov/~chao/uboone/docdb/files/LArProperty.pdf <https://www.phy.bnl.gov/~chao/uboone/docdb/files/LArProperty.pdf>

So it seems ScintYieldRatio is being *set* to match literature as if it it's S/T, but is *used* to calculate the fast and slow photons in SimEnergyDeposit.h as if it were = S/(S+T). I can see this being confusing down the line, as most users will be incorrectly setting this to be the singlet-to-triplet ratio, as seems to already be the case! 

May I suggest the term be changed to something more explicit, like “ScintSingletFraction”? Or more ideally, we could change it to “ScintSingletToTripletRatio” while also changing the behavior of the getter functions for SimEnergyDeposit to do the calculation properly:

// int NumFPhotons() const { return round(numPhotons * scintYieldRatio); }
// int NumSPhotons() const { return round(numPhotons * (1.0 - scintYieldRatio); }
int NumFPhotons() const { return round(numPhotons * scintSingletToTripletRatio/(1+scintSingletToTripletRatio) ); }
int NumSPhotons() const { return round(numPhotons * 1./(1.+scintSingletToTripletRatio) ); }

Thanks,
Will

aihimmel:
Yes, I think this is a good idea, but it's a little bit complicated since I think we will want to simultaneously update these functions, as well as the variables/functions/fhicl parameters in LArProperties and in the SimEnergyDeposit data product.

Do the LArSoft experts have any advice about how to coordinate something like this?

lgarren:
trigger build

lgarren:
@aihimmel  I think you just need a coordinated set of PR's.  Unless there is some "external" product that also needs to be updated.

lgarren:
@wforeman Whoops.  We have a c7 build failure:
https://dbweb5.fnal.gov:8443/LarCI/app/storage/docs/2020/05/26/buildLArSoft.log

Looks like just an unused variable.

lgarren:
Reminder, you can build with c7 on SLF7.

lgarren:
trigger build

lgarren:
approve

PR #23, opened 2020-06-04
Feature/plasorak decay0

plasorak:
OK... I've created yet another radiological generator (well actually 2):

The first one is making use of [Decay0](https://github.com/BxCppDev/bxdecay0). That's a software that is used in SuperNemo and a couple of other double beta decay experiments. It can generate radiactive decays which is what I use it for in this case. Right now, the libraries are in my home directory as you can see in the `CMakeLists.txt`. Hopefully this code can be added to UPS and we can link it directly? I don't know how to do that.

The second one is something that uses some of the `BaseRadioGen` to generate any spectrum of any particle in any volume at any rate (you can pass all these 4 in the fcl). In a sense, this is just a better version of `SingleGen` (except of course you're not getting 1 particle/event).

I've left the RadioGen_module intact (well... not its history...), such that people can use it if they want.

So, should this go ahead and get merged, these are the things that I think should be done:
- add some comments inside the code such that it can get picked up by doxygen (I don't really know how to do that).
- add CI tests, although I don't really know how to do that for function that generate object randomly or from a module.
- add `Decay0` in UPS.


tomjunk:
Sounds like we need a Decay0 UPS product, or subsume all of its code into an existing repository.  This creates a maintenance burden moving forwards.  What does Decay0 depend on?  I cannot see inside your home directory due to the new file permissions.  Perhaps copy it to /dune/data/users for inspection?  (app is more natural but it is too small). Decay0 would have to be updated when its authors make changes, and also when we move to new compilers or when a dependent product of it updates.  We do this with our big dependencies -- GENIE, GEANT, ROOT though it is a chore.

plasorak:
`find_package(GSL 2.4 REQUIRED)` is the only dependency that is advertised, as discussed in their github page (https://github.com/BxCppDev/bxdecay0):
_BxDecay0 is a standalone library with very few dependencies (mostly the GSL library for numerical integration and a few special math functions). External random engines can be used through a simple wrapping functor interface, particularly the ones provided by the C++ standard random library can be used by default. However the user is free to provide its own uniform deviates random generator (based on GSL, ROOT or whatever)._

I've copied the directory to `/dune/app/users/plasorak/bxdecay0` and checked the permissions (I think we can afford 50MB on dune/app?)

I understand the annoyance of adding a dependency. To be perfectly honest, I'm not entirely convinced myself. I've just already experienced what happens when we want to add a new beta decay in radiological generator: it takes time, we need to compute the spectrum ourselves, etc. The nice thing about this is we just change the fhicl. ¯\\\_(ツ)\_/¯

lgarren:
LArSoft has some concerns about the Decay0 license.  Some discussion may be forthcoming.

plasorak:
I just want to share this, that may help the discussion here: https://github.com/BxCppDev/bxdecay0/issues/5

fmauger:
> LArSoft has some concerns about the Decay0 license. Some discussion may be forthcoming.
Hi,

BxDecay0 makes use of the GSL library so the GPL License. That's the game with the FSF. No choice.
The same for the MathMore module in ROOT.
The only way I see to on-line bridge with a full software stack using proprietary or more permissive open source license is to provide a networked Client/Server interface with some possibly simple open exchange data format with proper licensing scheme. 

Another approach would use some file generation with a standalone program on BxDecay0's GPL side and a reader on the client app's side with another license. Fortunately, the "bxdecay0::event" class provides a "store" method which generate simple formatted ASCII output (https://github.com/BxCppDev/bxdecay0/blob/develop/bxdecay0/event.h).
A reader would be easy to implement on client's side. Of course such an approach implies some file management task but maybe not an issue for you.

Regards
frc

plasorak:
Hi all, is there a way to making this go forward? Or should we conclude that we are not going to use this?

lgarren:
We apologize for the delay.   Our requirements for any code using GPL v3 are that it be sequestered in a module.  We see that you have done this and that the code is in its own new directory. Thank you.  We will look at providing Decay0 as a ups product.  That is a required first step.

tomjunk:
I don't know if there's any problem with the GPL license.   LArSoft already depends on GSL.  I was more concerned about the need to maintain a new UPS product (Pierre's third bullet), which means building for all supported platforms (they are getting fewer.  DUNE plans to drop SL6 soon, and macOS is also getting dropped).  So that really means just SL7 (prof+debug)*(clang+gcc) = four builds.

plasorak:
I see, I just wanted to make sure that this is still going ahead. Thanks, and let me know if I can be of any help.

lgarren:
@tomjunk  The GPL v3 license is more restrictive than our apache license, and is therefore a concern.

sjmpeeters:
@tomjunk How should we proceed? The package is critical for the Background Task Force, as this will allow us to simulate important radioactive backgrounds correctly.

tomjunk:
I tried poking around the bxdecay0 package from github, and have a question about the data files of tabulated phase-space factors.  The README file says one needs a git add-on (LFS support) to obtain these files.  But beyond that, we should decide where to put them and how to point bxdecay0 to find them.  They are 1.7 GB big according to the README.rst file, but I haven't downloaded them yet.  We should put them in StashCache.  And make sure that the proper environment variable is set so a program can find these files and that they are not hardcoded.  bxdecay0-config has hardcoded directory names made when the package is built, but will break when relocated.  Small things to overcome.

tomjunk:
Oh I see after reading the docs a little more.  the 1.7 GB of files are only needed for the double-beta-decay processes of 82Se,  100Mo, 116Cd,  and 150Nd.  Probably not an issue for DUNE then to omit them.

fmauger:
Yes you are right, the "DBD gA" feature is very specific to the SuperNEMO experiment or possibly other projects which would be able to study fine DBD kinematics effects in the final state. So IMHO this simulation option is absolutely useless for DUNE so that you don't have to load these huge space factor files (however, when used, they are in principle downloaded from a public repo as an CMake ExternalProject during the build step and stored directly  in the bxdecay0 source directory before to be installed with other resource files)

Concerning a possible relocation of the package, it is possible to add some mechanism to guarantee this. I usually use the BinReloc stuff to ensure that with some success, whatever the software is made of, shared lib or executable. If you really need it I can add this feature.

plasorak:
Woops, sorry about that I didn't realise that decay0 was downloading huge DBD tables (well I did, but I then forgot). I've checked that everything works on my local install with the cmake option `-DBxDecay0_WITH_DBD_GA=OFF`.

lgarren:
Thanks for the information.  We are already working to make decay0 available as a ups package.

tomjunk:
Poking around in the source, I see that it depends on an environment variable, 
BXDECAY0_RESOURCE_DIR 

to be set to:

<install_directory>/share/BxDecay0-1.0.4/resources

though the test CMakeLists.txt file points to the version in the project source directory.  There are a few small files in ${BXDECAY0_RESOURCE_DIR}/data that it needs to read and run.

The version number in that directory name looks like it depends on the version of bxdecay0.  I had checked out the head of develop, and the last tagged version is 1.0.3.  So at the very least we should either fork the repo and tag it, or request a tag, or use version 1.0.3.

lgarren:
@tomjunk  In order for this to be part of larsoft, we need to supply an official decay0 package.  That work is in progress.  Anything you do at this time should be considered strictly temporary.  Also, the product will be named bxdecay0 to maintain the github "branding".

fmauger:
Concerning the use of the BXDECAY0_RESOURCE_DIR environment variable, it is optional.
The mechanism for locating resource files (only a few files of interest for the library if you don't use the DBD-gA generator which relies on a set of huge data files) is based on the installation path resolved at install step within CMake.
The  BXDECAY0_RESOURCE_DIR is made available in order to enable testing of an alternative resource file tree from an existing BxDecay0 installation. Such a task is usually only for expert developers of the package and probably not useful for you.
As BxDecay0 is not dynamically relocatable, I have just added the optional  BXDECAY0_PREFIX_DIR env var to be set manually in case one wants to move the full installation tree to some other place than the static one used at the CMake install step.
Tag 1.0.4 includes the change.

lgarren:
@fmauger  Thank you for the information.  Since all our ups products are relocated, we will definitely use the 1.0.4 tag.

fmauger:
Ok. That's nice if it is suitable for you. Do not hesitate to contact me or open an issue if you have a request or need help about bxdecay0. I do not expect big changes in the coming months but I will keep you informed if there is something hot about background generators that may be of some interest for DUNE.
Let me inform Vladimir Tretyak (author of the legacy F77+cernlib version) of your interest for our work. That's an additional motivation to maintain these tools and keep reactive.
Regards

sjmpeeters:
Thanks to all for your help in this. It's great that we will be able report this positive development in the background task force review next week!

fmauger:
Also we have considered more carefully this question of relocation. We will consolidate this aspect in next tag 1.0.5.
I don't know if you plan  to use CMake in your client code but the BxDecay0Config.cmake finder script that we will provide will define the "BxDecay0::BxDecay0" target to identify the libBxDecay0.so shared lib and associated material (all being relocatable within the same relative filesystem layout [usual bin, include, lib, share/... dirs from a single prefix]).
I'll keep you informed.

fmauger:
PS: Reference for "relocatable" support of BxDecay0: https://github.com/BxCppDev/bxdecay0/pull/8

fmauger:
https://github.com/BxCppDev/bxdecay0/releases/tag/1.0.5 is now available and should be fully relocatable if nothing is changed in the relative layout of the installation tree (bin/, include/, lib/, share/). 
Let me know if everything work on your side.

plasorak:
https://cdcvs.fnal.gov/redmine/projects/larsim/wiki/Running_Decay0_simulation_in_LArSim for documentation about the implementation in larsim.

lgarren:
bxdecay0 v1_0_5 is now available on larsoft cvmfs.  We feel that the interface to BxDecay0 is best separated into its own repository, larsimrad.  We have populated larsimrad with the changes in this PR.   This is analogous to the larrecodnn repository.  I think you will see more separation like this in future.  Please test larsimrad and make sure it works for you.

lgarren:
Forgot to mention that you will need mrb v4_01_03, which is just now being installed on the larsoft cvmfs.

plasorak:
Thanks Lynn!! I'll have a look tomorrow.

lgarren:
larsimrad is now part of the larsoft release distributuion, closing this PR

PR #27, opened 2020-07-03
Bug fix on LightSource

PetrilloAtWork:
I have found a bug I introduced with my last pull request.
Normally `LightSource` module produces each photon on a randomised location offset from a central location.
I was actually always writing the central location to the event, instead of the randomised one.
The fix is quite simple. Just pick the right variable, for once.


lgarren:
trigger build

lgarren:
Most experiment code is not yet updated, but the unit tests are fine.  Fast tracking this bug fix.

lgarren:
approve

PR #28, opened 2020-07-03
New Sim Filter using the sim::SimPhotonsLite data product

gputnam:
Hello!

This request incorporates a new filter into the `SimFilters/` directory. It is a module that replicates the functionality of the `FilterSimPhotonTime` filter except using `sim::SimPhotonsLite` instead of `sim::SimPhotons` as a data product.

I elected to write this as a new module (as opposed to porting `FilterSimPhotonTime` to handle both cases) because the code is minimal and mostly involves parsing the `sim::SimPhotonsLite` data product. Thus, porting `FilterSimPhotonTime` to handle both cases would mostly involve something like:

```
if (hasSimPhotons)
    FilterSimPhotons()
else
    FilterSimPhotonsLite()
```
and so I thought it would be cleaner to just write a new module. In addition, `FilterSimPhotonTime` requires cutting on the energy of the photons, which is not possible using the `sim::SimPhotonsLite` data product, and so the filter would have to be reorganized to handle both data products.

I have verified that the functionality works on a number of example input files.

I am not 100% sure on the procedure to merge content into larsoft, so please let me know what steps I should take in order to get this merged. Thank you!

### DIFF
--- a/larsim/SimFilters/FilterSimPhotonLiteTime_module.cc
+++ b/larsim/SimFilters/FilterSimPhotonLiteTime_module.cc
@@ -21,6 +21,7 @@
 #include <iostream>
 #include <string>
 #include <vector>
+#include <utility>
 
 #include "lardataobj/Simulation/SimPhotons.h"
 
@@ -43,7 +44,7 @@ private:
   bool filter(art::Event& e, art::ProcessingFrame const&) override;
 
   std::string const fSimPhotonsLiteCollectionLabel; //!< Label for the sim::SimPhotonsLite data product
-  std::vector<std::vector<int>> const fTimeWindows; //!< Time windows used for filtering. Units are the same as in the sim::SimPhotonsLite
+  std::vector<std::pair<int, int>> const fTimeWindows; //!< Time windows used for filtering. Units are the same as in the sim::SimPhotonsLite
   int const fMinTotalPhotons; //!< Minimum number of photons inside a window to pass the filter
   bool const fDebug; //!< Set to true to print (a lot of) debug information.
   std::size_t const fN; //!< Number of time winows.
@@ -58,7 +59,7 @@ simfilter::FilterSimPhotonLiteTime::FilterSimPhotonLiteTime(
   art::ProcessingFrame const&)
   : SharedFilter{p}
   , fSimPhotonsLiteCollectionLabel(p.get<std::string>("SimPhotonsLiteCollectionLabel"))
-  , fTimeWindows(p.get<std::vector<std::vector<int>>>("TimeWindows"))
+  , fTimeWindows(p.get<std::vector<std::pair<int, int>>>("TimeWindows"))
   , fMinTotalPhotons(p.get<int>("MinTotalPhotons"))
   , fDebug(p.get<bool>("Debug", false))
   , fN(fTimeWindows.size())
@@ -87,18 +88,13 @@ simfilter::FilterSimPhotonLiteTime::CheckTimeWindows() const
               << fTimeWindows.size() << std::endl;
 
   for (auto const& tw : fTimeWindows) {
-    if (tw.size() != 2)
-      throw cet::exception("FilterSimPhotonLiteTime::CheckTimeWindows")
-        << "Bad time window initialization: time window has wrong size (not 2)."
-        << std::endl;
-
     if (fDebug)
       std::cout << "\t\tTimeWindow "
-                << "[" << tw[0] << "," << tw[1] << "]" << std::endl;
+                << "[" << tw.first << "," << tw.second << "]" << std::endl;
 
-    if (tw[0] > tw[1])
+    if (tw.first > tw.second)
       throw cet::exception("FilterSimPhotonLiteTime::CheckTimeWindows")
-        << "Bad time window initialization: tw[0]>tw[1]. Reverse the order!"
+        << "Bad time window initialization: tw.first>tw.second. Reverse the order!"
         << std::endl;
   }
 }
@@ -130,9 +126,9 @@ simfilter::FilterSimPhotonLiteTime::filter(art::Event& e,
                 << simphotonslite.OpChannel << std::endl;
 
     for (auto const& photon_pair: simphotonslite.DetectedPhotons) {
-      for (size_t i_tw = 0; i_tw < fN; ++i_tw) {
+      for (size_t i_tw = 0; i_tw < fN; i_tw++) {
         auto const& tw(fTimeWindows[i_tw]);
-        if (photon_pair.first >= tw[0] && photon_pair.first <= tw[1]) {
+        if (photon_pair.first >= tw.first && photon_pair.first <= tw.second) {
 
           if (fDebug) {
             std::string photon_string = (i_pc < simPhotonsLiteCollection.size()) ? "Photon" : "Reflected Photon";
@@ -158,7 +154,7 @@ simfilter::FilterSimPhotonLiteTime::filter(art::Event& e,
               << fMinTotalPhotons << ":" << std::endl;
     for (size_t i_tw = 0; i_tw < fN; ++i_tw) {
       std::cout << "\t\tTimeWindow "
-                << "[" << fTimeWindows[i_tw][0] << "," << fTimeWindows[i_tw][1]
+                << "[" << fTimeWindows[i_tw].first << "," << fTimeWindows[i_tw].second
                 << "]: " << sumNPhotonArray[i_tw] << std::endl;
     }
   }
--- a/larsim/SimFilters/FilterSimPhotonLiteTime_module.cc
+++ b/larsim/SimFilters/FilterSimPhotonLiteTime_module.cc
@@ -4,6 +4,11 @@
 // File:        FilterSimPhotonLiteTime_module.cc
 //
 // Author: Gray Putnam -- ported from FilterSimPhotonTime
+//
+// Module for filtering events based on the number of true photons
+// hitting optical detectors inside a time window. Uses the
+// sim::SimPhotonsLite data product as input (see FilterSimPhotonTime
+// for filtering using the sim::SimPhotons data product).
 ////////////////////////////////////////////////////////////////////////
 
 #include "art/Framework/Core/ModuleMacros.h"
@@ -38,13 +43,13 @@ public:
 private:
   bool filter(art::Event& e, art::ProcessingFrame const&) override;
 
-  std::string const fSimPhotonsLiteCollectionLabel;
-  std::vector<std::vector<int>> const fTimeWindows;
-  int const fMinTotalPhotons;
-  bool const fDebug;
-  std::size_t const fN;
-  bool fUseReflectedPhotons;
-  std::string fReflectedLabel;
+  std::string const fSimPhotonsLiteCollectionLabel; //!< Label for the sim::SimPhotonsLite data product
+  std::vector<std::vector<int>> const fTimeWindows; //!< Time windows used for filtering. Units are the same as in the sim::SimPhotonsLite
+  int const fMinTotalPhotons; //!< Minimum number of photons inside a window to pass the filter
+  bool const fDebug; //!< Set to true to print (a lot of) debug information.
+  std::size_t const fN; //!< Number of time winows.
+  bool fUseReflectedPhotons; //!< Whether to include reflected photons in the filter.
+  std::string fReflectedLabel; //!< Label for the reflected photons -- "Reflected" by default.
 
   void CheckTimeWindows() const;
 };
--- a/larsim/SimFilters/FilterSimPhotonLiteTime_module.cc
+++ b/larsim/SimFilters/FilterSimPhotonLiteTime_module.cc
@@ -1,0 +1,165 @@
+////////////////////////////////////////////////////////////////////////
+// Class:       FilterSimPhotonLiteTime
+// Module Type: filter
+// File:        FilterSimPhotonLiteTime_module.cc
+//
+// Author: Gray Putnam -- ported from FilterSimPhotonTime
+////////////////////////////////////////////////////////////////////////
+
+#include "art/Framework/Core/ModuleMacros.h"
+#include "art/Framework/Core/SharedFilter.h"
+#include "art/Framework/Principal/Event.h"
+#include "art/Framework/Principal/Handle.h"
+#include "fhiclcpp/ParameterSet.h"
+
+#include <memory>
+#include <iostream>
+#include <string>
+#include <vector>
+
+#include "lardataobj/Simulation/SimPhotons.h"
+#include "lardataobj/Simulation/SimPhotons.h"
+
+namespace simfilter {
+  class FilterSimPhotonLiteTime;
+}
+
+class simfilter::FilterSimPhotonLiteTime : public art::SharedFilter {
+public:
+  explicit FilterSimPhotonLiteTime(fhicl::ParameterSet const& p,
+                               art::ProcessingFrame const&);
+
+  // Plugins should not be copied or assigned.
+  FilterSimPhotonLiteTime(FilterSimPhotonLiteTime const&) = delete;
+  FilterSimPhotonLiteTime(FilterSimPhotonLiteTime&&) = delete;
+  FilterSimPhotonLiteTime& operator=(FilterSimPhotonLiteTime const&) = delete;
+  FilterSimPhotonLiteTime& operator=(FilterSimPhotonLiteTime&&) = delete;
+
+private:
+  bool filter(art::Event& e, art::ProcessingFrame const&) override;
+
+  std::string const fSimPhotonsLiteCollectionLabel;
+  std::vector<std::vector<int>> const fTimeWindows;
+  int const fMinTotalPhotons;
+  bool const fDebug;
+  std::size_t const fN;
+  bool fUseReflectedPhotons;
+  std::string fReflectedLabel;
+
+  void CheckTimeWindows() const;
+};
+
+simfilter::FilterSimPhotonLiteTime::FilterSimPhotonLiteTime(
+  fhicl::ParameterSet const& p,
+  art::ProcessingFrame const&)
+  : SharedFilter{p}
+  , fSimPhotonsLiteCollectionLabel(p.get<std::string>("SimPhotonsLiteCollectionLabel"))
+  , fTimeWindows(p.get<std::vector<std::vector<int>>>("TimeWindows"))
+  , fMinTotalPhotons(p.get<int>("MinTotalPhotons"))
+  , fDebug(p.get<bool>("Debug", false))
+  , fN(fTimeWindows.size())
+  , fUseReflectedPhotons(p.get<bool>("UseReflectedPhotons", false))
+  , fReflectedLabel(p.get<std::string>("fReflectedLabel", "Reflected"))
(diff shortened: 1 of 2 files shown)


knoepfel:
@gputnam, thanks for the PR!  Your approach in making this a separate module is sensible; and it's threadsafe!  One could think of fancy ways to extend the current filter to be able to handle the `SimPhotonsLite` situation.  But that may be more difficult than it's worth.

Although this is largely a cut-and-paste of the current `FilterSimPhotonTime`, are you willing to consider some refactoring that could improve the layout of this code?  For example, trying to iterate over two photon collections in the same for loop is awkward.  The following would be a cleaner design:

```{.cpp}
std::vector<int> sumNPhotonArray(fN, 0);
if (overThreshold(photons, sumNPhotonArray)) return true;
if (overThreshold(reflectedPhotons, sumNPhotonArray)) return true;
return false;
```

I realize this reorganization was not in the scope of what you were intending to implement, but it would help clarify the intention of the module.

gputnam:
Thank you for the reply!

On the suggested change -- I don't think it would work as is. This is because the photons and reflected photons could combine to get an event over the threshold of the specified number of minimum PE's. So one couldn't check if they go over threshold individually.

I agree the code could be organized better -- but it would be my preference to keep the organization of the SmPhotons and SimPhotonsLite filter to be nearly identical. Because otherwise it would be non-obvious to a reader/user that they indeed do (basically) the same thing.

knoepfel:
@gputnam, keeping the organization the same between SimPhotons and SimPhotonsLite is reasonable.  Hopefully down the road we can improve it, but I'm willing to let this go in for now.  I'll approve the PR once we get L2 approval.

ghost:
trigger build

aihimmel:
While simultaneously supporting SimPhotons and SimPhotonsLite is probably more trouble than it's worth, you might consider also support OpDetBacktrackerRecords. They were designed as a replacement for SimPhotonsLite (in DUNE at least we've moved away from SimPhotonsLite for some time), support all the same functionality, plus keep enough information to allow truth backtracking. If you're game, I can dig up some examples for you.

ghost:
CI tests are OK.

larsim/SimFilters/FilterSimPhotonLiteTime_module.cc
```diff
@@ -0,0 +1,170 @@
...
+private:
+  bool filter(art::Event& e, art::ProcessingFrame const&) override;
+
+  std::string const fSimPhotonsLiteCollectionLabel; //!< Label for the sim::SimPhotonsLite data product
+  std::vector<std::vector<int>> const fTimeWindows; //!< Time windows used for filtering. Units are the same as in the sim::SimPhotonsLite
+  int const fMinTotalPhotons; //!< Minimum number of photons inside a window to pass the filter
+  bool const fDebug; //!< Set to true to print (a lot of) debug information.
+  std::size_t const fN; //!< Number of time winows.
+  bool fUseReflectedPhotons; //!< Whether to include reflected photons in the filter.
+  std::string fReflectedLabel; //!< Label for the reflected photons -- "Reflected" by default.
```
absolution1: Could these be const as well?

larsim/SimFilters/FilterSimPhotonLiteTime_module.cc
```diff
@@ -0,0 +1,170 @@
...
+  fhicl::ParameterSet const& p,
+  art::ProcessingFrame const&)
+  : SharedFilter{p}
+  , fSimPhotonsLiteCollectionLabel(p.get<std::string>("SimPhotonsLiteCollectionLabel"))
+  , fTimeWindows(p.get<std::vector<std::vector<int>>>("TimeWindows"))
+  , fMinTotalPhotons(p.get<int>("MinTotalPhotons"))
+  , fDebug(p.get<bool>("Debug", false))
+  , fN(fTimeWindows.size())
+  , fUseReflectedPhotons(p.get<bool>("UseReflectedPhotons", false))
+  , fReflectedLabel(p.get<std::string>("fReflectedLabel", "Reflected"))
```
absolution1: Default values can be a bit of a ticking time bomb e.g. if someone mis-spells 'fReflectedLabel' as 'ReflectedLabel', they will end up using "Reflected" as the label and my not realise it.
Are the defaults needed?

I'm guessing that 'fReflectedLabel' should be 'ReflectedLabel'?

larsim/SimFilters/FilterSimPhotonLiteTime_module.cc
```diff
@@ -0,0 +1,170 @@
...
+simfilter::FilterSimPhotonLiteTime::filter(art::Event& e,
+                                       art::ProcessingFrame const&)
+{
+  auto const& simPhotonsLiteCollection =
+    *e.getValidHandle<std::vector<sim::SimPhotonsLite>>(fSimPhotonsLiteCollectionLabel);
+
+  std::vector<int> sumNPhotonArray(fN, 0);
+
+  const std::vector<sim::SimPhotonsLite> &simPhotonsLiteCollectionReflected = fUseReflectedPhotons ?
+    *e.getValidHandle<std::vector<sim::SimPhotonsLite>>({fSimPhotonsLiteCollectionLabel, fReflectedLabel}) : std::vector<sim::SimPhotonsLite>();
```
absolution1, line 112: Just curious, is grabbing `std::vector<sim::SimPhotonsLite> &` instead of `std::vector<art::Ptr<sim::SimPhotonsLite>>` a thread-safety thing?

larsim/SimFilters/FilterSimPhotonLiteTime_module.cc
```diff
@@ -0,0 +1,170 @@
...
+#include "art/Framework/Principal/Handle.h"
+#include "fhiclcpp/ParameterSet.h"
+
+#include <memory>
+#include <iostream>
+#include <string>
+#include <vector>
+
+#include "lardataobj/Simulation/SimPhotons.h"
+#include "lardataobj/Simulation/SimPhotons.h"
```
absolution1: Duplicate header

larsim/SimFilters/FilterSimPhotonLiteTime_module.cc
```diff
@@ -0,0 +1,170 @@
...
+  const std::vector<sim::SimPhotonsLite> &simPhotonsLiteCollectionReflected = fUseReflectedPhotons ?
+    *e.getValidHandle<std::vector<sim::SimPhotonsLite>>({fSimPhotonsLiteCollectionLabel, fReflectedLabel}) : std::vector<sim::SimPhotonsLite>();
+
+  size_t n_sim_photons = simPhotonsLiteCollection.size() + simPhotonsLiteCollectionReflected.size();
+
+  if (fDebug) {
+    std::cout << "New event to filter with total # sim photons: " << n_sim_photons << std::endl;
+  }
+    
+  for (size_t i_pc = 0; i_pc < n_sim_photons; i_pc++) {
```
absolution1, line 120: --really really-- minor, you do `i_pc++` here but `++i_tw` below.

larsim/SimFilters/FilterSimPhotonLiteTime_module.cc
```diff
@@ -0,0 +1,170 @@
...
+        << std::endl;
+
+    if (fDebug)
+      std::cout << "\t\tTimeWindow "
+                << "[" << tw[0] << "," << tw[1] << "]" << std::endl;
+
+    if (tw[0] > tw[1])
+      throw cet::exception("FilterSimPhotonLiteTime::CheckTimeWindows")
+        << "Bad time window initialization: tw[0]>tw[1]. Reverse the order!"
+        << std::endl;
```
absolution1: Seeing as you're being extra careful with the inputs, perhaps you should defend against the two time windows being the same?

larsim/SimFilters/FilterSimPhotonLiteTime_module.cc
```diff
@@ -0,0 +1,170 @@
...
+
+  void CheckTimeWindows() const;
+};
+
+simfilter::FilterSimPhotonLiteTime::FilterSimPhotonLiteTime(
+  fhicl::ParameterSet const& p,
+  art::ProcessingFrame const&)
+  : SharedFilter{p}
+  , fSimPhotonsLiteCollectionLabel(p.get<std::string>("SimPhotonsLiteCollectionLabel"))
+  , fTimeWindows(p.get<std::vector<std::vector<int>>>("TimeWindows"))
```
absolution1: Based on the below code, it looks like each vector element --needs-- to be length two.  You could naturally enforce this by using `std::vector<std::pair<float,float> >`

absolution1 (commented):
Hi Gray.
Really nice stuff 👍 
I've added a few really minor comments besides what others have mentioned that you're very much free to pick/reject as you wish.

larsim/SimFilters/FilterSimPhotonLiteTime_module.cc
knoepfel: @absolution1, a few things:
- The data product is `std::vector<sim::SimPhotonsLite>`, not `std::vector<art::Ptr<sim::SimPhotonsLite>>`.  In general, `art::Ptr`s should only be used when needing to persist a product reference to disk.
- If the `simPhotonsLiteCollectionReflected` variable were not a reference, then a potentially expensive copy operation would occur by calling `std::vector`'s copy constructor.
- Thread-safety is not an issue here as the framework guarantees that `getValidHandle` is thread-safe, and the `simPhotonsLiteCollectionReflected` variable is immutable (by being marked `const`)

absolution1: Hi @knoepfel 
Thanks for the info.
I was specifically asking why art::Ptr wasn't used, as that would avoid having to use a reference at all.  Using the vector filled with `art::Ptr<PhotonLite>` also wouldn't have the same hefty copy cost, and it would mean storing a reference to a vector could be avoided.
Aren't the `sim::SimPhotonsLite` coming from disk?  If so, I'm not sure I understand why the `art::Ptr` isn't available to use.

At least in SBND's case, the `sim::SimPhotonsLite` are stored in the same way as the other data products, which are regularly used via `art::Ptrs`:

G4.......... | largeant...... | ..................... | std::vector<simb::MCParticle>........................................ | ..11
G4.......... | largeant...... | ..................... | std::vector<sim::AuxDetSimChannel>................................... | 2272
G4.......... | mcreco........ | ..................... | std::vector<sim::MCTrack>............................................ | ...1
G4.......... | largeant...... | Reflected............ | std::vector<sim::OpDetBacktrackerRecord>............................. | .193
G4.......... | largeant...... | ..................... | art::Assns<simb::MCTruth,simb::MCParticle,sim::GeneratedParticleInfo> | ..11
G4.......... | largeant...... | Reflected............ | std::vector<sim::SimPhotonsLite>..................................... | .193
G4.......... | largeant...... | ..................... | std::vector<sim::SimChannel>......................................... | .680
G4.......... | largeant...... | ..................... | std::vector<sim::SimPhotonsLite>..................................... | .202

knoepfel: @absolution1, there appear to be a few misunderstandings:

- Because `art::Ptr`s are designed for persistency, they are three times bigger than a regular pointer.  Large vectors of `art::Ptr`s are thus expensive in terms of memory and I/O.  There has been an over-reliance on them in LArSoft-based code, and we'd like to clean that up.
- The `std::vector<sim::SimPhotonsLite>` type is the type of the data product, either in memory or on disk.  There is no seamless translation from that to `std::vector<art::Ptr<sim::SimPhotonsLite>>`.  Bringing about such a translation requires creating `art::Ptr`s, which in itself is expensive, and does not happen automatically.
- Even if the data product in question could be translated to `std::vector<art::Ptr<sim::SimPhotonsLite>>`, that has nothing to do with whether `simPhotonsLiteCollectionReflected` is an object or a reference:
    ```.cpp
    std::vector<sim::SimPhotonsLite> const& coll = 
      *e.getValidHandle(...); // obj is a reference to the product, no copy
    std::vector<art::Ptr<SimPhotonsLite>> coll = 
      translate(*e.getValidHandle(...)); // coll is owned by value, increasing memory footprint of job
    ```
It's almost always preferable to own objects by value and not by pointer.

absolution1: @knoepfel 

I'm not sure that there's been so much of a misunderstanding as I just don't think I'm explaining myself very well.  I was not definitely not aware how cumbersome the `art::Ptr` is, though.

If LArSoft optimisation is at the point where art::Ptrs are considered hefty then, fine, I can completely see why one would avoid using them.
However, `art::Ptrs` offer a certain degree of protection which, to the best of my knowledge, a reference to a vector does not.
If any of the `art::Ptrs` in the vector go stale and are subsequently accessed then an exception is thrown and everything is bailed gracefully, right?
If the reference to the vector goes stale and is accessed, there would be a segfault?

For sure I was not aware that the `art::Ptr` was considered so hefty these days.
I'm curious about how much memory and CPU time is lost by doing this instead storing the underlying vector as a reference?

```
art::Handle<std::vector<sim::SimPhotonsLite> > photons;
e.getByLabel(label, photons);
for (size_t iPhoton = 0; iPhoton < photons.size(); ++iPhoton){
  const art::Ptr<sim::SimPhotonLite> photon(photons, iPhoton);
//stuff to do with the photon
}
```

For your last point, I think I've caused a lot of confusion; I was not in any way suggesting that storing by value vs reference are equivalent in terms of CPU/memory cost. It wasn't even my intent to suggest storing an entire `std::vector<art::Ptr<simb::SimPhotonsLite>>` (one could convert each element of the handle as needed instead using the above snippet), although I see that my very first comment certainly did imply it.  However, seeing as it's been brought up, I've always thought that constructing the `art::Ptrs` and saving them in a new vector vs taking a reference to the entire vector can be preferable, because of the modest global difference in CPU/memory cost and also the safety reasons I mentioned above.  But, I was basing everything on the assumption that art::Ptrs are cheap to construct/copy/move, which it doesn't sound like they are...

gputnam:
@absolution1 @knoepfel -- regarding the above comments. The intended scope of this pull request was to introduce a new module that was more-or-less a copy of the SimPhotonTime filter except using the sim::SimPhotonsLite data product. 

The above comments apply equally to the SimPhotonTime filter. And, as I stated earlier, from my perspective it would be best if the two filters are kept nearly identical to best communicate their paired functionality.

Since I'm already looking at these two filters, I would be happy to apply the above comments to both the SimPhotonTime and SimPhotonTimeLite filters. @absolution1 -- I definitely agree with most of the suggestions. However, since the CI tests have already passed my preference would be to merge this pull request and then open a new pull request with the further suggested changes.

What do people think?

larsim/SimFilters/FilterSimPhotonLiteTime_module.cc
knoepfel: @absolution1, you've brought up issues that are worth discussing, but at this point it is a discursion from the PR.  The lifetime of the reference in question is localized to the `produce` function call.  Unless the code is caching the reference across `produce` calls (it's not), there will be no segmentation violation...at least not from using a reference.

The calculation of the overhead for a `vector<art::Ptr<T>>` object is not hard:

| Variable | Memory usage (bytes) | Notes |
| -------- | ------------- | ------- |
| `vector<T> const&` | 8 | Size of pointer |
| `vector<art::Ptr<T>>` | 24 x (N + 1) | N is the number of vector elements, and the extra 1 is due to how `std::vector` is implemented |

Even an empty vector owned by value is more expensive than just the reference to an already-owned object.

We should continue further discussion elsewhere, but in short, the reference is just fine in this context.

gputnam:
Alternatively, we could also just expand the scope of this pull request.

larsim/SimFilters/FilterSimPhotonLiteTime_module.cc
aihimmel: This is a function of history. Support for reflected photons was added to lots of existing modules, and we didn't want to force breaking fhicl changes on the whole LArSoft world.

absolution1:
Hi @gputnam 
All of my comments are minor/optional and you should totally feel free to pick/choose/ignore whichever you like (maybe drop the duplicate header though!).

A second PR is fine in principle but in practice such touch-up PRs are easily forgotten/lost.  I'd suggest that, if any of changes are going to be adopted, it's probably best to apply them as part of this PR.

gputnam:
Ok! Then I'll add in some commits that change both Filters to address these comments.

larsim/SimFilters/FilterSimPhotonLiteTime_module.cc
gputnam: Fixed

larsim/SimFilters/FilterSimPhotonLiteTime_module.cc
gputnam: I kept these as having defaults, but I did change "fReflectedLabel" to "ReflectedLabel"

larsim/SimFilters/FilterSimPhotonLiteTime_module.cc
gputnam: Fixed.

larsim/SimFilters/FilterSimPhotonLiteTime_module.cc
gputnam: Fixed.

larsim/SimFilters/FilterSimPhotonLiteTime_module.cc
gputnam: I don't think it's best to check this -- I agree such a configuration would be weird, but it wouldn't be strictly incorrect or cause wrong behavior.

larsim/SimFilters/FilterSimPhotonLiteTime_module.cc
gputnam: Fixed.

gputnam:
Just pushed some commits addressing the above comments

lgarren:
trigger build

absolution1:
hi @gputnam 
Thanks for opting to make those small amendments, I took a quick look and thought they all looked good.  I think I noticed bit of mixing between `i_tw++` and `++i_tw` but I wouldn't worry about that...
I must admit that I did not expect:
```
fTimeWindows(p.get<std::vector<std::pair<int, int>>>("TimeWindows"))
```
to parse so nicely, good stuff!

I don't have any further comments but I won't put the approval signature in as I saw there are currently other conversations going on.

absolution1 (approved):

knoepfel:
trigger build

lgarren:
CI tests are normal
This PR is ready for level 2 approval

absolution1:
I'm not sure if there was an ongoing convo between @gputnam and @aihimmel.  I was about to message @aihimmel on slack but I think he's currently ill.
Based on the brief convo between them above, it sounds like such a change could be implemented down the line.
So, I'm happy to approve.  Please revert the approval if you wish.

absolution1:
approve

lgarren:
approve

aihimmel:
Agreed. I was requesting an addition more than a change, so it can come later. It could also be imagined as a broader move from SimPhotonsLite to OpDetBacktrackerRecords as the primary data product used.

PR #29, opened 2020-07-09
Prevent implicit downcasting from double to int in VUVHits() and VISH…

ikatza:
…its()

This is meant to address issue #24433

### DIFF
--- a/larsim/LegacyLArG4/OpFastScintillation.cxx
+++ b/larsim/LegacyLArG4/OpFastScintillation.cxx
@@ -493,9 +493,6 @@ namespace larg4 {
     // get the number of photons produced from the IonizationAndScintillation
     // singleton
     larg4::IonizationAndScintillation::Instance()->Reset(&aStep);
-    // TODO: I believe MeanNumberOfPhotons should be unsigned int or
-    // long unsigned int, this would need to change across different files
-    // ~icaza
     double MeanNumberOfPhotons = larg4::IonizationAndScintillation::Instance()->NumberScintillationPhotons();
     // double stepEnergy          = larg4::IonizationAndScintillation::Instance()->VisibleEnergyDeposit()/CLHEP::MeV;
     RecordPhotonsProduced(aStep, MeanNumberOfPhotons);//, stepEnergy);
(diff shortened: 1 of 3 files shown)


knoepfel:
@ikatza, thanks for the PR.  This bit of code can always use some cleaning up, but you have appropriately limited yourself to addressing the `double`-to-`int` implicit conversions.  The PR will receive L1 approval.

PetrilloAtWork:
There are chances that this pull request might conflict with #30, just in terms or (GIT) merging and not in terms of functionality.
If that turns out to be the case, I am available to help with the merging.

ghost:
trigger build

aihimmel:
@ikatza, it looks like the _inputs_ have been changed to doubles, but the _outputs_ have been left as integers. From our discussion, I thought the plan was to make everything floating point during the entire G4 process, and leave the Poisson throws an integers until the last moment before waveforms get created.

ikatza:
Hi @aihimmel, I couldn't really figure out a way to do it that wouldn't involve a lot of work. Nor I'm convinced that what I've come up would work. 

The output of those functions represent hits, these quantities are passed to a routine that assigns a propagation time to each hit. A collection of `Photon`s or a `LitePhoton`s are created at this stage, the amount of them is precisely these integers that get passed.

I wouldn't go about trying to change that behaviour.

The approach that I thought might be implemented is to pass doubles, still create collections with integer amount of photons, but the last one representing a fraction of a photon. So if I get 7.7 vuvhits, then there will be 7 regular `Photons` or `LitePhotons`, and then one extra that is only 0.7 of a photon.

Obviously this is not currently implemented and it's not what is expected. For instance `AddLitePhoton` looks like:

```void OpDetPhotonTable::AddLitePhoton( int opchannel, int time, int nphotons, bool Reflected)```

Whilst it's a simple change to swap `int nphotons` to `double nphotons` it has a number of implications further down. 

The non-lite variety doesn't allow any meaningful way to add fractions of photons. You are expected to create a `sim::OnePhoton`
```
sim::OnePhoton PhotToAdd;
PhotToAdd.InitialPosition  = PhotonPosition;
PhotToAdd.Energy           = PhotonEnergy;
PhotToAdd.Time             = Time;
PhotToAdd.SetInSD          = false;
PhotToAdd.MotherTrackID    = tracknumber;
```

aihimmel:
OK, fair enough. In most circumstances if QE hasn't been applied yet the number of photons being made is sufficiently large that it should matter too much. Thanks for checking.

aihimmel:
Approve

aihimmel (approved):

lgarren:
approve

PR #30, opened 2020-07-10
Proposed resolution for issue #24600.

PetrilloAtWork:
This is the first version of the proposed solution to [Redmine issue #24600](https://cdcvs.fnal.gov/redmine/issues/24600).

There are two main assumptions that have been applied in general, but should have been applied only when using the semi-analytic model (or better, when using it with SBND detector and maybe ProtoDUNE, for which those assumptions are satisfied):

1. scintillation light is propagated only from the TPC active volume
    * for the way this is implemented, another assumption is required: the detector has a single monolithic active volume, e.g. it has only one cryostat
2. scintillation light does not go through the cathode, which is fully opaque

My pull request is made of several commits that can be grouped in three sets:

1. preparation (commits 731a89f109ac8afed4eb4e134ba466210ecb4309, 84ca9f375d793ec161b163ccccd2a5b7ce60e215, e2af37560b0f7e5ca994b1524fc526830d04fff6, 4b743969a88bf844817819d7208b86559fb7914b and 23f5d8011126b19578fead727122309f55a1bd8a):
    * the determination of the active volumes is corrected to extract them from all cryostats; LArSoft geometry facilities are used for this;
      _the requirement above is not lifted_: this is a change in preparation for the figure support of multi-cryostat detectors
    * `OpFastScintillation` object is copiable but the copy is bugged (both copy will attempt to destroy the same memory); I turned it into a moveable, non-copiable object, as copy was not necessary (nor is move, actually)
    * introduced an explicitly named `OpFastScintillation::usesSemiAnalyticModel()` method that will help to highlight the model-specific parts of the code
    * these changes do not change the behaviour of the code
2. fix (commit 1ab8718061dcac30182564c2435d3475c92d9fe9):
    * introduced `fOnlyActiveVolume` flag marking the implementation of light propagation only from the active volume; if this flag is not set, the implementation will avoid steps which would require light propagation only from the active volume; in other words, if `fOnlyActiveVolume` is `true` if the scintillation source is not in the active volume it is not propagated, if it is `false` the propagation is attempted no matter where the source is;
 this is automatically set to `true` only if the semi-analytic model is used (`usesSemiAnalyticModel()` is `true`)
    * introduced `fOnlyOneCryostat` option: setting this to `true` will acknowledge that the photon propagation _might_ not be performed in cryostats beyond the first one (`C:0`); if this option is `false`, the detector has more than one cryostat and semi-analytic mode is used, an exception will be thrown, otherwise run will proceed;
      this is meant as a temporary option for _testing_ the semianalytic model on a single cryostat in detectors like ICARUS, while the support to the full detector is being developed by somebody else; this flag is currently hard-coded to `false`, i.e. using semi-analytic model on multi-cryostat detectors will always throw an exception, while other models are unaffected
    * introduced `fOpaqueCathode` flag marking the code assuming cathode opaqueness; if this flag is `true`, optimisations may be enabled not to propagate the scintillation light to optical detectors that are not facing the same TPC where scintillation happens; otherwise, all optical detectors are considered for propagation (any TPC, any cryostat); this flag is currently hard-coded to `false` because LArSoft does not have a way to detect nor express whether the cathode is transparent or opaque; this is for use in future implementations.
3. some internals interacting with `GeometryCore` service provider have been converted to use `geo::Point_t` as point type instead of `std::array<double, 3>`, since the former is the recommended LArSoft interface (commits 9e07e5ae3de9ce31e5908b993f4c71e77e4d830f and bfe915b7c560113d9d4a79008fbab064488a7c3e); this change does not affect the behaviour of the class and it is not required in the resolution of the issue.

It can be noted that the "fixes" at point 2 are very little about actions and a lot about marking assumptions throughout the code.
The hope is that this will make a bit easier to have full support of semi-analytic, which the SBN program is studying as an option, model for detectors like ICARUS. That will be [Redmine issue #24601](https://cdcvs.fnal.gov/redmine/issues/24601)

I request review of these changes by @ikatza who is one of the developers of the affected code.

### DIFF
--- a/larsim/LegacyLArG4/OpFastScintillation.cxx
+++ b/larsim/LegacyLArG4/OpFastScintillation.cxx
@@ -220,7 +220,7 @@ namespace larg4 {
           << ")"
           ;
       } // local scope
-      
+
       if (usesSemiAnalyticModel() && (geom.Ncryostats() > 1U)) {
         if (fOnlyOneCryostat) {
           mf::LogWarning("OpFastScintillation")
@@ -303,7 +303,7 @@ namespace larg4 {
       if(usesSemiAnalyticModel()) {
         mf::LogVerbatim("OpFastScintillation")
           << "OpFastScintillation: using semi-analytic model for number of hits";
-        
+
         // LAr absorption length in cm
         std::map<double, double> abs_length_spectrum = lar::providerFrom<detinfo::LArPropertiesService>()->AbsLengthSpectrum();
         std::vector<double> x_v, y_v;
@@ -1505,13 +1505,13 @@ namespace larg4 {
     }
   }
 
-  
+
   // ---------------------------------------------------------------------------
   bool OpFastScintillation::usesSemiAnalyticModel() const {
     return fUseNhitsModel;
   } // OpFastScintillation::usesSemiAnalyticModel()
-  
-  
+
+
   // ---------------------------------------------------------------------------
   void OpFastScintillation::detectedDirectHits(std::map<size_t, int>& DetectedNum,
                                                const double Num,
@@ -2107,29 +2107,29 @@ namespace larg4 {
     return 0.;
   }
 
-  
+
   // ---------------------------------------------------------------------------
   std::vector<geo::BoxBoundedGeo> OpFastScintillation::extractActiveVolumes
     (geo::GeometryCore const& geom)
   {
     std::vector<geo::BoxBoundedGeo> activeVolumes;
     activeVolumes.reserve(geom.Ncryostats());
-    
+
     for (geo::CryostatGeo const& cryo: geom.IterateCryostats()) {
-      
+
       // can't use it default-constructed since it would always include origin
       geo::BoxBoundedGeo box { cryo.TPC(0).ActiveBoundingBox() };
-      
+
       for (geo::TPCGeo const& TPC: cryo.IterateTPCs())
         box.ExtendToInclude(TPC.ActiveBoundingBox());
-        
+
       activeVolumes.push_back(std::move(box));
-      
+
     } // for cryostats
-    
+
     return activeVolumes;
   } // OpFastScintillation::extractActiveVolumes()
-  
+
   // ---------------------------------------------------------------------------
 
 
--- a/larsim/LegacyLArG4/OpFastScintillation.cxx
+++ b/larsim/LegacyLArG4/OpFastScintillation.cxx
@@ -349,23 +349,6 @@ namespace larg4 {
   }
 
 
-  OpFastScintillation::OpFastScintillation(const OpFastScintillation& rhs)
-    :  G4VRestDiscreteProcess(rhs.GetProcessName(), rhs.GetProcessType())
-  {
-    theSlowIntegralTable        = rhs.GetSlowIntegralTable();
-    theFastIntegralTable        = rhs.GetFastIntegralTable();
-
-    fTrackSecondariesFirst      = rhs.GetTrackSecondariesFirst();
-    fFiniteRiseTime             = rhs.GetFiniteRiseTime();
-    YieldFactor                 = rhs.GetScintillationYieldFactor();
-    ExcitationRatio             = rhs.GetScintillationExcitationRatio();
-    scintillationByParticleType = rhs.GetScintillationByParticleType();
-    emSaturation                = rhs.GetSaturation();
-
-    BuildThePhysicsTable();
-  }
-
-
   ////////////////
   // Destructors
   ////////////////
--- a/larsim/LegacyLArG4/OpFastScintillation.cxx
+++ b/larsim/LegacyLArG4/OpFastScintillation.cxx
@@ -178,6 +178,8 @@ namespace larg4 {
     , bPropagate(!(art::ServiceHandle<sim::LArG4Parameters const>()->NoPhotonPropagation()))
     , fPVS(bPropagate? art::ServiceHandle<phot::PhotonVisibilityService const>().get(): nullptr)
     , fUseNhitsModel(fPVS && fPVS->UseNhitsModel())
+      // for now, limit to the active volume only if semi-analytic model is used
+    , fOnlyActiveVolume(usesSemiAnalyticModel())
   {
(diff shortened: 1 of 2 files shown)


ikatza:
Hi. Thanks for these changes Gianluca. The assumptions are made clear and there's less clutter. I have inspected the commits and they all seems good.

I've run it on sbndcode and confirm that we still get direct and reflected light. Light from scintillation coming from one TPC doesn't transport to another, as expected. And light is being correctly assigned to the TPC it originated.

To this last point, I hit a snag because the convention for TPC numbering on the event display that comes up with `evd_sbnd.fcl` is opposite to the convention chosen for the optical channels and TPC associated to those detectors. I'll follow up on that on sbndcode. 

These changes set the way to address [issue 24601](https://cdcvs.fnal.gov/redmine/issues/24601)

There are merge conflicts trying to fuse with PR #29, not hard to solve.

I can provide the merge resolution for that PR and finish moving away from `std::array<double, 3>` to `geo::Point_t` as this PR started.

Other than I prefer to write `const type` and not `type const`, I have no further comments and from my point of view it can be merged.

larsim/LegacyLArG4/OpFastScintillation.cxx
```diff
@@ -890,8 +883,8 @@ namespace larg4 {
     G4int numOfMaterials = G4Material::GetNumberOfMaterials();
 
     // create new physics table
-    if(!theFastIntegralTable)theFastIntegralTable = new G4PhysicsTable(numOfMaterials);
-    if(!theSlowIntegralTable)theSlowIntegralTable = new G4PhysicsTable(numOfMaterials);
+    if(!theFastIntegralTable)theFastIntegralTable = std::make_unique<G4PhysicsTable>(numOfMaterials);
+    if(!theSlowIntegralTable)theSlowIntegralTable = std::make_unique<G4PhysicsTable>(numOfMaterials);
```
knoepfel, line 885: Thank you!

larsim/LegacyLArG4/OpFastScintillation.cxx
```diff
@@ -1059,36 +1052,36 @@ namespace larg4 {
...
     }
-    else if (pvs->IncludePropTime()) {
+    else if (fPVS->IncludePropTime()) {
       // Get VUV photons arrival time distribution from the parametrization
-      const G4ThreeVector OpDetPoint(fOpDetCenter.at(OpChannel)[0]*CLHEP::cm,
-                                     fOpDetCenter.at(OpChannel)[1]*CLHEP::cm,
-                                     fOpDetCenter.at(OpChannel)[2]*CLHEP::cm);
+      const G4ThreeVector OpDetPoint(fOpDetCenter.at(OpChannel).X()*CLHEP::cm,
+                                     fOpDetCenter.at(OpChannel).Y()*CLHEP::cm,
+                                     fOpDetCenter.at(OpChannel).Z()*CLHEP::cm);
```
knoepfel: Clearer, yes.  For efficiency reasons, I'd prefer something like:
```.cpp
auto const& point = fOpDetCenter.at(OpChannel);
const G4ThreeVector OpDetPoint(point.X()*CLHEP::cm,
                               point.Y()*CLHEP::cm,
                               point.Z()*CLHEP::cm);
```
No reason to check the range more than once--and I know you didn't introduce the multiple  `at(...)` calls.

larsim/LegacyLArG4/OpFastScintillation.cxx
```diff
@@ -1059,36 +1052,36 @@ namespace larg4 {
...
         getVUVTimes(arrival_time_dist, distance_in_cm); // in ns
       }
       else {
         TVector3 ScintPoint( x0[0]/CLHEP::cm, x0[1]/CLHEP::cm, x0[2]/CLHEP::cm ); // in cm
-        TVector3 OpDetPoint_tv3(fOpDetCenter.at(OpChannel)[0],
-                                fOpDetCenter.at(OpChannel)[1],
-                                fOpDetCenter.at(OpChannel)[2]); // in cm
+        TVector3 OpDetPoint_tv3(fOpDetCenter.at(OpChannel).X(),
+                                fOpDetCenter.at(OpChannel).Y(),
+                                fOpDetCenter.at(OpChannel).Z()); // in cm
```
knoepfel: Same comments as above.

knoepfel (commented):
Looks good, Gianluca.  Just one minor comment.

aihimmel:
Just to make sure all is clear, there was a concern raised on the Redmine issue if this might negatively impact protoDUNE by changing the behavior of the parameterization. The answer is that I think the changes here should be a non-issue for protoDUNE since 
1) The cathode is opaque
2) ProtoDUNE is using photon libraries, and probably will be fore some time.

ghost:
trigger build

lgarren:
@PetrilloAtWork @ikatza  We have merged larsim PR 29 with develop.  We hope that will enable you to fix the merge conflicts on this branch.

larsim/LegacyLArG4/OpFastScintillation.cxx
PetrilloAtWork: I overlooked a lot of good practice issues... including this one.
Since I have to merge it again, I'll fix this as well.

PetrilloAtWork:
Thank you. ETA 2 hours :crossed_fingers:

PetrilloAtWork:
I have merged the current `larsim` `develop` into my feature branch, and took the chance to implement the change suggested by Kyle (or something along those lines).
It should be good now.

PetrilloAtWork:
(also, I apologise that I have seen only now the comment of @ikatza about the `type const` vs. `const type`... and I slipped a few of the former in)

lgarren:
trigger build

lgarren:
I have started the CI builds, but be aware that dcache is down.

aihimmel:
The changes to the code look sensible by eye, but perhaps we should wait for the CI tests to be rerun tomorrow?

lgarren:
trigger build

lgarren:
trigger build

once more with feeling

lgarren:
dcache still down, after talking to Erica, we are moving ahead

lgarren:
approve

PR #31, opened 2020-07-18
Features to speed up photon library generation [1/2]

PetrilloAtWork:
I have added some features to the legacy `LArG4` chain to speed up the generation of a photon visibility library.
The changes in `larsim` include:
* option not to produce `simb::MCParticle` collection; this yield a major memory saving since a photon library job may generate hundred of thousand scintillation-like photons;
* option to use `sim::SimPhotonsLite` instead of `sim::SimPhotons`, which would save individual photons; the information we require for the photon library is just the total number of detected photons, and possibly a time distribution;
* as a "bonus" I have added support for some metadata in the photon library: number of voxels and volume covered.

To be valuable, these features need to be combined with improvements in `SimPhotonCounters` module, which is the target of the twin [pull request #8](https://github.com/LArSoft/larana/pull/8) for `larana`.

As a technical note: metadata is implemented via writing `RooInt` and `RooDouble` objects into the `TFileService` ROOT file: I could not find any other `TObject`-like object which would contain a single value and play well with `TFileService` registry interface (it was not obvious to me how to write non-`TObject` objects like `TVectorD`, neither.

### DIFF
--- a/larsim/LegacyLArG4/OpDetSensitiveDetector.cxx
+++ b/larsim/LegacyLArG4/OpDetSensitiveDetector.cxx
@@ -16,13 +16,16 @@
 #include "larsim/LegacyLArG4/OpDetPhotonTable.h"
 #include "larsim/LegacyLArG4/OpDetLookup.h"
 #include "lardataobj/Simulation/SimPhotons.h"
+#include "larcoreobj/SimpleTypesAndConstants/PhysicalConstants.h" // util::pi()
 #include "Geant4/G4SDManager.hh"
 
 namespace larg4{
 
 
-  OpDetSensitiveDetector::OpDetSensitiveDetector(G4String DetectorUniqueName)
+  OpDetSensitiveDetector::OpDetSensitiveDetector
+    (G4String DetectorUniqueName, bool useLitePhotons /* = false */)
     : G4VSensitiveDetector(DetectorUniqueName)
+    , fUseLitePhotons(useLitePhotons)
   {
     // Register self with sensitive detector manager
     G4SDManager::GetSDMpointer()->AddNewDetector(this);
@@ -36,28 +39,42 @@ namespace larg4{
 
   //--------------------------------------------------------
 
-  G4bool OpDetSensitiveDetector::ProcessHits(G4Step * aStep, G4TouchableHistory *)
+  void OpDetSensitiveDetector::AddLitePhoton(G4Step const* aStep, int OpDet)
+  {
+    
+    double const time  = aStep->GetTrack()->GetGlobalTime();
+
+    // the guideline: if it's VUV (~128 nm) is direct, otherwise it is reflected
+    double const energy = aStep->GetTrack()->GetVertexKineticEnergy() / CLHEP::eV;
+    bool const reflected = Wavelength(energy) > 200.0; // nm
+    
+    // Add this photon to the detected photons table
+    fThePhotonTable->AddLitePhoton(OpDet, static_cast<int>(time), 1, reflected);
+    
+  } // OpDetSensitiveDetector::AddLitePhoton()
+
+  
+  //--------------------------------------------------------
+
+  void OpDetSensitiveDetector::AddPhoton(G4Step const* aStep, int OpDet)
   {
     sim::OnePhoton ThePhoton;
-
 
     // Get photon data to store in the hit
 
     ThePhoton.SetInSD      = true;
 
-    auto const& startPos = aStep->GetTrack()->GetVertexPosition();
+    auto const& track = *(aStep->GetTrack());
+    auto const& startPos = track.GetVertexPosition();
     ThePhoton.InitialPosition = { startPos.x(), startPos.y(), startPos.z() };
 
-    //ThePhoton.Time                = aStep->GetTrack()->GetGlobalTime() - fGlobalTimeOffset;
-    ThePhoton.Time                = aStep->GetTrack()->GetGlobalTime();
+    //ThePhoton.Time                = track.GetGlobalTime() - fGlobalTimeOffset;
+    ThePhoton.Time                = track.GetGlobalTime();
 
-
-    ThePhoton.Energy              =  aStep->GetTrack()->GetVertexKineticEnergy();
+    ThePhoton.Energy              =  track.GetVertexKineticEnergy();
 
     // Lookup which OpDet we are in
-    G4StepPoint *preStepPoint = aStep->GetPreStepPoint();
-
-    int OpDet = fTheOpDetLookup->GetOpDet(preStepPoint->GetPhysicalVolume());
+    G4StepPoint const* preStepPoint = aStep->GetPreStepPoint();
 
     // Store relative position on the photon detector
     G4ThreeVector worldPosition  = preStepPoint->GetPosition();
@@ -68,9 +85,22 @@ namespace larg4{
       localPosition.z()/CLHEP::cm
       };
 
-
     // Add this photon to the detected photons table
     fThePhotonTable->AddPhoton(OpDet, std::move(ThePhoton));
+
+  } // OpDetSensitiveDetector::AddPhoton()
+
+  //--------------------------------------------------------
+
+  G4bool OpDetSensitiveDetector::ProcessHits(G4Step * aStep, G4TouchableHistory *)
+  {
+    // Lookup which OpDet we are in
+    int const OpDet
+      = fTheOpDetLookup->GetOpDet(aStep->GetPreStepPoint()->GetPhysicalVolume());
+
+    // Add this photon to the detected photons table
+    if (fUseLitePhotons) AddLitePhoton(aStep, OpDet);
+    else                 AddPhoton(aStep, OpDet);
 
     // Kill this photon track
     aStep->GetTrack()->SetTrackStatus(fStopAndKill);
@@ -86,5 +116,17 @@ namespace larg4{
   {
 
   }
+  
+  
+  //--------------------------------------------------------
+  constexpr double OpDetSensitiveDetector::Wavelength(double energy) {
+    
+    // SI 2019 (eV nm):
+    constexpr double hc = 6.62607015e-34 * 299792458.0 / 1.602176634e-19 * 1e9;
+    
+    return hc / energy; // nm
+    
+  } // OpDetSensitiveDetector::Wavelength()
+  
 
 }
--- a/larsim/LegacyLArG4/OpDetSensitiveDetector.cxx
+++ b/larsim/LegacyLArG4/OpDetSensitiveDetector.cxx
@@ -19,6 +19,14 @@
 #include "larcoreobj/SimpleTypesAndConstants/PhysicalConstants.h" // util::pi()
 #include "Geant4/G4SDManager.hh"
 
+namespace {
+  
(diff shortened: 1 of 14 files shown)


ghost:
trigger build

ghost:
CI tests are normal.

ghost:
approve

larsim/LegacyLArG4/OpDetSensitiveDetector.h
```diff
@@ -58,10 +58,24 @@ namespace larg4 {
...
     //double                     fGlobalTimeOffset;
+    
+    /// Adds the photon at the specified step with full information.
+    void AddPhoton(G4Step const* aStep, int OpDet);
+    
+    /// Adds the photon at the specified step with reduced information.
+    void AddLitePhoton(G4Step const* aStep, int OpDet);
+    
+    /// Converts a photon `energy` [eV] into its Wavelength [nm]
+    static constexpr double Wavelength(double energy);
```
knoepfel: As this function is private and not intended to be exposed, I suggested moving it into the implementation file as a free function in an anonymous namespace.

larsim/LegacyLArG4/OpDetSensitiveDetector.cxx
```diff
@@ -36,28 +39,42 @@ namespace larg4{
...
     ThePhoton.InitialPosition = { startPos.x(), startPos.y(), startPos.z() };
 
-    //ThePhoton.Time                = aStep->GetTrack()->GetGlobalTime() - fGlobalTimeOffset;
-    ThePhoton.Time                = aStep->GetTrack()->GetGlobalTime();
-
+    //ThePhoton.Time                = track.GetGlobalTime() - fGlobalTimeOffset;
+    ThePhoton.Time                = track.GetGlobalTime();
 
-    ThePhoton.Energy              =  aStep->GetTrack()->GetVertexKineticEnergy();
+    ThePhoton.Energy              =  track.GetVertexKineticEnergy();
```
knoepfel, line 82: Can `ThePhoton` be initialized at its point of declaration?

larsim/LegacyLArG4/ParticleListAction.h
```diff
@@ -47,7 +48,7 @@ namespace larg4 {
 
     struct ParticleInfo_t {
 
-      simb::MCParticle* particle = nullptr;  ///< simple structure representing particle
+      simb::MCParticle* particle = nullptr;  ///< Object representing particle (not owned).
```
knoepfel: Suggest using `cet::exempt_ptr` to make the ownership semantics part of the type.

larsim/LegacyLArG4/ParticleListAction.h
```diff
@@ -105,6 +109,9 @@ namespace larg4 {
     std::map<int, GeneratedParticleIndex_t> const& GetPrimaryTruthMap() const
       { return fPrimaryTruthMap; }
 
+    /// Returns whether a particle list is being kept.
+    bool hasList() const { return bool(fparticleList); }
```
knoepfel: Recommend:
```.cpp
bool hasList() const noexcept { return static_cast<bool>(fparticleList); }
```

larsim/LegacyLArG4/ParticleListAction.h
```diff
@@ -124,7 +131,7 @@ namespace larg4 {
                                                      ///< be included in the list.
     ParticleInfo_t           fCurrentParticle;       ///< information about the particle currently being simulated
                                                      ///< for a single particle.
-    sim::ParticleList*       fparticleList;          ///< The accumulated particle information for
+    std::unique_ptr<sim::ParticleList> fparticleList; ///< The accumulated particle information for
```
knoepfel, line 135: This is an improvement, thanks.

larsim/PhotonPropagation/PhotonLibrary.cxx
```diff
@@ -12,16 +12,41 @@
...
+  
+  template <typename RooT, typename T>
+  struct RooReader {
+    TDirectory& srcDir;
+    std::vector<std::string>& missingKeys;
+    
+    std::optional<T> operator() (std::string const& key)
+      {
+        RooT const* value = srcDir.Get<RooT>(key.c_str());
+        if (value) return { T(*value) };
```
knoepfel: Suggest:
```.cpp
if (value) return std::make_optional<T>(*value);
```

larsim/PhotonPropagation/PhotonLibrary.cxx
```diff
@@ -398,6 +438,127 @@ namespace phot{
...
+      zMin, zMax, zN
+      );
+    
+  } // PhotonLibrary::LoadMetadata()
+  
+  
+  //------------------------------------------------------------
+  void PhotonLibrary::StoreMetadata() const {
+    
+    auto const& tfs = *(art::ServiceHandle<art::TFileService const>());
```
knoepfel: My preference would be for the `art::TFileService` reference to be passed in instead of creating a handle inside a file that is not intrinsically related to art.  The pattern you've adopted is consistent with what has been done before, but LArSoft needs to be getting away from it.

larsim/PhotonPropagation/PhotonLibrary.h
```diff
@@ -65,7 +72,26 @@ namespace phot{
...
+    // --- BEGIN --- Metadata: voxel information -------------------------------
+    /// @name Metadata: voxel information
+    /// @{
+    
+    /// Returns whether voxel metadata is available.
+    bool hasVoxelDef() const { return fVoxelDef.has_value(); }
+
+    /// Returns the current voxel metadata (undefined behaviour if none).
+    /// @see `hasVoxelDef()`
+    sim::PhotonVoxelDef const& GetVoxelDef() const { return fVoxelDef.value(); }
```
knoepfel: Actually the behavior is well-defined in this case: an exception will be thrown if the optional object contains no value.  Perhaps you'd like to do the following instead?
```.cpp
sim::PhotonVoxelDef const& GetVoxelDef() const noexcept 
{
  assert(hasVoxelDef());
  return *fVoxelDef;
}
```

knoepfel (commented):
Thanks for the PR, Gianluca.   A few comments for your consideration.

aihimmel:
Photon library metadata! This has been a **major** pain point with the photon libraries for many years. Thank you for putting this together, @PetrilloAtWork.

aihimmel:
A quick read of the code changes looks OK to me. I will approve from the physics side while you keep working on the technical comments from Kyle.

aihimmel:
Approve

larsim/LegacyLArG4/OpDetSensitiveDetector.h
PetrilloAtWork: Ok. Apparently I somehow have a bias against free functions.

larsim/LegacyLArG4/OpDetSensitiveDetector.cxx
PetrilloAtWork: It is, with an aggregate initialization.
I try to avoid that (until C++20) because a change in the definition may mess up the initialization (in this case, if for any reason `Energy` and `Time` are swapped).
I suggest we leave it this way, but I let you have the last word on the topic.

larsim/LegacyLArG4/ParticleListAction.h
PetrilloAtWork: I appreciate the intention. If I read the code, though, I am lead to think that `particle` pointer is not entitled to overtime work, rather than about data ownership. So the difference is between having all readers read the documentation on the same line of the data member, and having all non-SciSoft readers search for the `exempt_ptr.h` header to discover the same information. I think that in "modern" code a bare pointer is customary not to own its pointee... is that wrong?
As I said, I appreciate the intention though, and I will oblige.

larsim/LegacyLArG4/ParticleListAction.h
PetrilloAtWork: I miss something more descriptive...

larsim/PhotonPropagation/PhotonLibrary.h
PetrilloAtWork: Yes, the behaviour of the current implementation is predictable, but I don't want to create expectations.
I accept the suggested implementation.

larsim/PhotonPropagation/PhotonLibrary.cxx
PetrilloAtWork: Ok.

larsim/LegacyLArG4/ParticleListAction.h
knoepfel: Two issues:

- C-style casts are are discouraged
- The `std::unique<T>::operator bool()` member function is guaranteed to not throw an exception, therefore the `noexcept` keyword.

larsim/LegacyLArG4/OpDetSensitiveDetector.h
PetrilloAtWork: Commit 0336a38.

larsim/LegacyLArG4/ParticleListAction.h
PetrilloAtWork: Commit 941efe1.

larsim/LegacyLArG4/ParticleListAction.h
PetrilloAtWork: Sorry, I meant that I miss an interface element of `std::unique_ptr` more descriptive than a cast to a different type.
Like `cet::exempt_ptr::empty()`, for example.

PetrilloAtWork: Commit e554efd.

larsim/PhotonPropagation/PhotonLibrary.cxx
PetrilloAtWork: Commit b54de44.

larsim/PhotonPropagation/PhotonLibrary.cxx
PetrilloAtWork: Ok, I will humour you, with the understanding that I consider `PhotonVisibilityService` a lost battle.
I have moved _art_ services out of `PhotonLibrary` entirely. It will be passed in by the caller, if the caller thinks it's needed.
And if the caller thinks wrong, they'll get an exception at the right time.
In `PhotonVisibilityService`, `TFileService` is required if building the library, optional if there is a request not to load the library, and skipped otherwise.

PetrilloAtWork: Commits fcbd706 and ed7e0ac.

larsim/PhotonPropagation/PhotonLibrary.h
PetrilloAtWork: Commit c497e35.

larsim/LegacyLArG4/OpDetSensitiveDetector.cxx
knoepfel: It's a minor inefficiency but not worth holding up the PR.

knoepfel:
approve

knoepfel:
trigger build

knoepfel:
trigger build with pull requests #31, larana#8

lgarren:
CI tests are normal.  This PR will be merged by hand during the release tagging process.

PR #32, opened 2020-07-20
Add (hopefully) simple to use truth matching utilities.

absolution1:
Utilities use the back tracker service under the hood.

All are freestanding functions wrapped in a namespace, rather than in a class structure.

The functions take either a `art::Ptr<recob::Hit>` or a `std::vector<art::Ptr<recob::Hit> >` and return the ID of the most significantly contributing G4 particle.  There are different functions for the various definitions of 'most significantly contributing'.

I've attempted to balance efficiency and readability:  for utility functions that tend to act on containers that only have ~10 or so elements in them (the elements here are G4 IDs matched to the hit vectors), I think a certain amount of inefficiency can be allowed if it helps with clarity/intent.

I am very very open to opinions/suggestions/criticisms on efficiency/readability/style/whatever else :)


ghost:
trigger build

ghost:
CI tests are normal.

ghost:
approve

knoepfel:
Thanks for the PR, @absolution1.   Do you know of places where code can be cleaned up by using these utilities?

absolution1:
Hi @knoepfel 
Truth matching has always been a faction kind of thing, so I think there's probably quite a few places in the various experiment's code where they do things 'their own way', which I'd --really-- like to unify.
 - I'd need to dig out places in `dunetpc`
 - I --think-- that ProtoDUNE run their own truth matching routines, but I'd need to confirm this.
 - `sbndcode` has been using these utilities for awhile (though I bet there's one or two cases where SBND collaborators have still done their own thing). 
 - I believe that larpandora uses its own truth matching routine to match hits to true particles.

I am sure there are others.  A well placed `grep BackTracker` might find many more cases...

larsim/Utils/TruthMatchUtils.cc
```diff
@@ -0,0 +1,178 @@
...
+
+namespace
+{
+
+using namespace TruthMatchUtils;
+IDToEDepositMap::const_iterator MaxEDepElementInMap(const IDToEDepositMap &idToEDepMap)
+{
+    IDToEDepositMap::const_iterator highestContribIt(std::max_element(idToEDepMap.begin(), idToEDepMap.end(),
+        [](const std::pair<G4ID, EDeposit>& a, const std::pair<G4ID, EDeposit>& b)->bool{ return
+        std::nextafter(a.second, std::numeric_limits<EDeposit>::lowest()) <= b.second && std::nextafter(a.second, std::numeric_limits<EDeposit>::max()) >= b.second
```
knoepfel, line 32: What is the reason for using `std::nextafter`?

larsim/Utils/TruthMatchUtils.cc
```diff
@@ -0,0 +1,178 @@
...
+} // namespace
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+TruthMatchUtils::G4ID TruthMatchUtils::TrueParticleID(const art::Ptr<recob::Hit> &pHit, const bool rollupUnsavedIDs)
+{
+    IDToEDepositMap idToEDepMap;
+    TruthMatchUtils::FillG4IDToEnergyDepositMap(idToEDepMap, pHit, rollupUnsavedIDs);
+    if (0 == idToEDepMap.size())
```
knoepfel: Recommend:
```.cpp
if (idToEDepMap.empty())
```
instead.  And likewise in various places below.

larsim/Utils/TruthMatchUtils.cc
```diff
@@ -0,0 +1,178 @@
...
+    const bool rollupUnsavedIDs)
+{
+    IDToEDepositMap idToChargeDepMap;
+    for (const art::Ptr<recob::Hit> &pHit : pHits)
+    {
+        const G4ID g4ID(TruthMatchUtils::TrueParticleID(pHit, rollupUnsavedIDs));
+        const EDeposit recoCharge(static_cast<EDeposit>(pHit->Integral()));
+        std::pair<IDToEDepositMap::iterator, const bool> insertPair(idToChargeDepMap.insert(IDToEDepositMap::value_type(g4ID, recoCharge)));
+        if (!(insertPair.second))
+            insertPair.first->second += recoCharge;
```
knoepfel: This can be written more clearly as:
```.cpp
auto const [iterator, inserted] = idToChargeDepMap.try_emplace(g4ID, recoCharge);
if (not inserted) {
  iterator->second += recoCharge;
}
```

larsim/Utils/TruthMatchUtils.cc
```diff
@@ -0,0 +1,178 @@
...
+TruthMatchUtils::G4ID TruthMatchUtils::TrueParticleIDFromTotalRecoHits(const std::vector<art::Ptr<recob::Hit> >& pHits,
+    const bool rollupUnsavedIDs)
+{
+    std::map<G4ID, unsigned int> idToHitCountMap;
+    for (const art::Ptr<recob::Hit> &pHit : pHits)
+    {
+        const G4ID g4ID(TruthMatchUtils::TrueParticleID(pHit, rollupUnsavedIDs));
+        std::pair<std::map<G4ID, unsigned int>::iterator, const bool> insertPair(idToHitCountMap.insert(std::map<G4ID, unsigned int>::value_type(g4ID, 1)));
+        if (!(insertPair.second))
+            insertPair.first->second++;
```
knoepfel: See above regarding more concise notation.

larsim/Utils/TruthMatchUtils.cc
```diff
@@ -0,0 +1,178 @@
...
+    if (0 == idToHitCountMap.size())
+        return kNoG4ID;
+
+    std::map<unsigned int, std::vector<G4ID> > hitCountToIDMap;
+    for (std::map<G4ID, unsigned int>::const_iterator mapIt = idToHitCountMap.begin(); mapIt != idToHitCountMap.end(); ++mapIt)
+    {
+        const G4ID g4ID(mapIt->first);
+        const unsigned int hitCount(mapIt->second);
+        std::pair<std::map<unsigned int, std::vector<G4ID> >::iterator, const bool> insertPair(
+            hitCountToIDMap.insert(std::map<unsigned int, std::vector<G4ID> >::value_type(hitCount, std::vector<G4ID>(1, g4ID))));
```
knoepfel: Recommend more concise form:
```.cpp
for (auto const& [g4ID, hitCount] : idToHitCountMap) {
  auto const [iterator, inserted] = hitCountToIDMap.try_emplace(hitCount, {g4ID});
  ...
}
```

larsim/Utils/TruthMatchUtils.h
```diff
@@ -0,0 +1,80 @@
...
+#include <limits>
+#include <vector>
+
+namespace TruthMatchUtils {
+
+typedef int G4ID;
+typedef float EDeposit;
+typedef std::map<G4ID, EDeposit> IDToEDepositMap;
+
+const G4ID kNoG4ID = std::numeric_limits<G4ID>::lowest();           ///< The value used when no G4 ID has been found
```
knoepfel: Generally not a good idea to put global variables in header files due to multiple definitions of the same variable.  If you have to do this, use an inline `constexpr` variable:
```.cpp
inline constexpr auto kNoG4ID = std::numeric_limits<G4ID>::lowest();
```

knoepfel (commented):
@absolution1, a few comments for your consideration primarily related to using more modern C++ facilities.  

At the moment, it sounds as if there may not be concrete use cases yet for these facilities.  Because of that, I'm a bit leery going forward knowing that any code added to LArSoft needs to be maintained.  Nonetheless, let's not let that concern prevent this PR from proceeding, especially as there is L2  approval.

absolution1:
Hi @knoepfel 
Thanks so much for the comments.

Regarding use cases, apologies but I completely forgot to mention ICARUS which also uses them.  I do already have at least one use case in dunetpc (my own code).  For all cases where the functions are in use, the header/imp are copied to the relevant experiment code, which makes maintaining a pain.
I'd also previously spoken to someone else in pandora about the desire to unify the truth matching between larsoft and pandora, so I think the use case is definitely there.

larsim/Utils/TruthMatchUtils.cc
absolution1: It's for equality checking `a` and `b`.  I went down a rabbit hole when finding the preferred way for equality-checking floating point values...

larsim/Utils/TruthMatchUtils.h
absolution1: There was a reason for doing this.  
The functions can return `kNoG4ID` when no G4 ID is found.  I figured it would be handy for end users to check if the function returned a sensible G4ID by comparing to `kNoG4ID`.  I also figured that as its wrapped in the namespace, the chance of a clash is minimal (provided users don't do something like `using namespace TruthMatchUtils` which I probably shouldn't assume...).

Is a `constexpr` data member still fine in this case or should I do something like return a literal from a function e.g.
```
inline constexpr G4ID TruthMatchUtils::NoG4ID() { return std::numeric_limits<G4ID>::lowest(); };
```

knoepfel: I suspected you had a reason for using the `kNoG4ID` constant.  If, however, the purpose of it is to test validity, it would be better to create an actual function in the header file that people can call (e.g.):

```.cpp
namespace TruthMatchUtils {
  bool valid(G4ID id) noexcept;
  ...
}
```

and then in the implementation file (.cc):

```.cpp
namespace {
  // Implementation detail, local to file
  constexpr auto invalid_id = std::numeric_limits<G4ID>::lowest();
}

namespace TruthMatchUtils {
  bool valid(G4ID const id) noexcept {
    return invalid_id != id;
  }
  ...
}
```

This way, users of the header simply do something like:
```.cpp
auto const id = TrueParticleIDFromTotalRecoCharge(...);
if (valid(id)) {
  // do something
}
```
which is clearer than an explicit comparison like:
```.cpp
if (id != kG4NoID) {
  // do something
}
```

larsim/Utils/TruthMatchUtils.cc
absolution1: Nice one.
I've updated all container size == 0 checks to use `empty()` instead.

larsim/Utils/TruthMatchUtils.cc
absolution1: I've opted to use the structured bindings as suggested throughout, but I do have some reservations about them.  I could not find a way in which I can declare some of the variables as const e.g. `auto [iterator, const inserted]`.  The scope of the vars isn't very long so its not much of a problem but it does slightly break the style...

larsim/Utils/TruthMatchUtils.cc
absolution1: I've picked up using structured bindings for the for loop and the return from `try_emplace`, to try and keep a consistent style throughout.
Besides what I said above about the structured bindings above, I do have one other reservation; I'm not a massive fan of auto, as I think it breaks one of c++'s main advantages (strong typing).  I'm probably in the minority here though...

larsim/Utils/TruthMatchUtils.h
absolution1: Totally agree, good idea.  I've implemented this.

larsim/Utils/TruthMatchUtils.cc
absolution1: Implemented.

absolution1:
Hi @knoepfel (I think you may be on holiday this week?) I've implemented all suggestions, though I do have a few minor readability reservations about the structured bindings (see above), but those reservations are not significant enough to avoid implementation.

absolution1:
trigger build

lgarren:
CI tests appear to be normal

ghost:
approve

lgarren:
approve

larsim/Utils/TruthMatchUtils.cc
knoepfel: There are many opinions about `auto`.  However, C++ does not lose strong-typing because of it--the compiler simply calculates what the type is...and that type cannot change for a given name.  It makes sense to use `auto` in small scopes; however once the scopes get too large, then it can be difficult to understand what it refers to.

absolution1: Sorry, I was using terminology incorrectly.  I was basically trying to say that auto makes the code harder to follow and I usually shy away from it.  With structured bindings, it seems like my hand is forced.

PR #33, opened 2020-07-22
Bug fix for  the factor of 2 difference between the original NEST alg

Healthborn:
two issues have been identified:
1. Unit of two variables
    Electric field: kV -> 1000 V
    Step length:    cm
2. Value of one variables
    Density of LAr

### DIFF
--- a/larsim/IonizationScintillation/ISCalcNESTLAr.cxx
+++ b/larsim/IonizationScintillation/ISCalcNESTLAr.cxx
@@ -57,53 +57,33 @@ namespace larg4
         }
 
         int pdgcode = edep.PdgCode();
-
-        geo::Length_t startx = edep.StartX();
-        geo::Length_t starty = edep.StartY();
-        geo::Length_t startz = edep.StartZ();
-        geo::Length_t endx   = edep.EndX();
-        geo::Length_t endy   = edep.EndY();
-        geo::Length_t endz   = edep.EndZ();
-//        geo::Point_t  midpos = edep.MidPoint();
-
-//        double delta    = 1*CLHEP::mm;
-//        double R0           = 1.568*CLHEP::um;    //Mozumder 1995
-
+        
         double DokeBirks[3];
-//        double ThomasImel = 0.00;
 
-        double eField = EFieldAtStep(fDetProp->Efield(), edep);
+//        double eField = EFieldAtStep(fDetProp->Efield(), edep);
+        double eField = EFieldAtStep(fDetProp->Efield(), edep) * 1000;  //conver kV to 1000 V
         if(eField)
         {
-//            ThomasImel   = 0.156977 * pow(eField, -0.1);
             DokeBirks[0] = 0.07 * pow((eField/1.0e3), -0.85);
             DokeBirks[2] = 0.00;
         }
         else
         {
-//            ThomasImel   = 0.099;
             DokeBirks[0] = 0.0003;
             DokeBirks[2] = 0.75;
         }
 
-        double Density = fDetProp->Density()/(CLHEP::g/CLHEP::cm3); // argon density at the temperature from Temperature()
+//        double Density = fDetProp->Density()/(CLHEP::g/CLHEP::cm3); // argon density at the temperature from Temperature()
+        double Density = LAr_Density;
 
-        //biExc = 0.6;
 
         // nuclear recoil quenching "L" factor: total yield is
         // reduced for nuclear recoil as per Lindhard theory
         double epsilon = 11.5 *(fEnergyDeposit/CLHEP::keV) * pow(LAr_Z,(-7./3.));
-//      double gamma   = 3. * pow(epsilon, 0.15) + 0.7 * pow( epsilon, 0.6 ) + epsilon;
-//      double kappa   = 0.133*pow(z1,(2./3.))*pow(a2,(-1./2.))*(2./3.);
 
         if ( pdgcode == 2112 || pdgcode == -2112 ) //nuclear recoil
         {
-//            fYieldFactor = (kappa * gamma) / ( 1 + kappa * gamma ); //Lindhard factor
             fYieldFactor = 0.23 * ( 1 + exp( -5 * epsilon ) ); //liquid argon L_eff
-//            if ( eField == 0 )
-//            {
-//                ThomasImel = 0.25; //special TIB parameters for nuclear recoil only, in LAr
-//            }
             fExcitationRatio = 0.69337 + 0.3065 * exp( -0.008806 *pow( eField, 0.76313 ) );
         }
 
@@ -168,11 +148,13 @@ namespace larg4
         }
         else //normal case of an e-/+ energy deposition recorded by Geant
         {
-            dx = std::sqrt((startx-endx)*(startx-endx)+(starty-endy)*(starty-endy)+(startz-endz)*(startz-endz))/CLHEP::cm;
+//            dx = std::sqrt((startx-endx)*(startx-endx)+(starty-endy)*(starty-endy)+(startz-endz)*(startz-endz))/CLHEP::cm;
+            dx = edep.StepLength();
             if(dx)
             {
                 LET = (dE/dx)*(1/Density); //lin. energy xfer (prop. to dE/dx)
             }
+            
             if ( LET > 0 && dE > 0 && dx > 0 )
             {
                 double ratio = CalcElectronLET( dE*1e3 )/LET;
@@ -295,7 +277,7 @@ namespace larg4
                               +(efield*eFieldOffsets.Z()*efield*eFieldOffsets.Z()) );
         }
 
-        return EField;
+        return EField; //unit: kV/cm
     }
 
     //......................................................................
--- a/larsim/IonizationScintillation/ISCalcNESTLAr.h
+++ b/larsim/IonizationScintillation/ISCalcNESTLAr.h
@@ -30,6 +30,7 @@
 #define LAr_Z       (18)
 #define Density_LAr (1.393)
 #define LAr_W_Value (19.5*CLHEP::eV)
+#define LAr_Density (1.4)
 
 namespace larg4
 {
--- a/larsim/PhotonPropagation/PDFastSimPAR_module.cc
+++ b/larsim/PhotonPropagation/PDFastSimPAR_module.cc
@@ -36,7 +36,6 @@
 #include "canvas/Utilities/InputTag.h"
 #include "nurandom/RandomUtils/NuRandomService.h"
 
-
 #include "fhiclcpp/ParameterSet.h"
 
 // LArSoft libraries
@@ -48,12 +47,11 @@
 #include "larsim/PhotonPropagation/PhotonVisibilityService.h"
 #include "larsim/PhotonPropagation/PhotonVisibilityTypes.h" // phot::MappedT0s_t
 #include "larsim/PhotonPropagation/ScintTimeTools/ScintTime.h"
-
+#include "larsim/Simulation/LArG4Parameters.h"
 #include "larsim/Simulation/LArG4Parameters.h"
 #include "larcore/Geometry/Geometry.h"
 #include "larcorealg/Geometry/CryostatGeo.h"
 #include "larcorealg/Geometry/OpDetGeo.h"
-
 
 // Random number engine
 #include "CLHEP/Random/RandFlat.h"
@@ -462,8 +460,19 @@ namespace phot
         std::cout << "PDFastSimPAR Module Construct" << std::endl;
 
         Initialization();
-        produces< std::vector<sim::SimPhotonsLite> >("par");
-        produces< std::vector<sim::OpDetBacktrackerRecord> >("par");
+        
+        art::ServiceHandle<sim::LArG4Parameters const> lgp;
+        if (lgp->UseLitePhotons())
+        {
+            std::cout << "Use Lite Photon." << std::endl;
+            produces< std::vector<sim::SimPhotonsLite> >("par");
+            produces< std::vector<sim::OpDetBacktrackerRecord> >("par");
+        }
+        else
+        {
+            std::cout << "Use Sim Photon." << std::endl;
+            produces< std::vector<sim::SimPhotons> >("par");
+        }
     }
 
     //......................................................................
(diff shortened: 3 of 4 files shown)


absolution1:
trigger build

lgarren:
@Healthborn @absolution1  would you take a look at the dune CI failure?  It may be unrelated to this PR, but I would like someone to take a look.

absolution1:
@lgarren it looks like an input file access issue again.

```
1372: %MSG-s ArtException:  TriggerResultInserter:TriggerResults@Construction  22-Jul-2020 14:53:15 CDT ModuleConstruction
1373: cet::exception caught in art
1374: ---- FileOpenError BEGIN
1375:   RootInputFileSequence::initFile(): Input file xroot://fndca1.fnal.gov:1095/pnfs/fnal.gov/usr/dune/persistent/stash/ContinuousIntegration/protoDUNEsp/datareco/np04_raw_run005809_0004_dl6.root was not found or could not be opened.
1376: ---- FileOpenError END
1377: %MSG
1378: 29.94user 2.95system 0:38.71elapsed 84%CPU (0avgtext+0avgdata 551804maxresident)k
1379: 1096464inputs+496outputs (3854major+305535minor)pagefaults 0swaps
1380: TIME SIGNATURE
1381: 187.60user 18.23system 7:00.41elapsed 48%CPU (0avgtext+0avgdata 551892maxresident)k
1382: 6538424inputs+3192outputs (25714major+2173039minor)pagefaults 0swaps
```

lgarren:
@absolution1  Thanks.  does this PR have level 2 approval?

larsim/IonizationScintillation/ISCalcNESTLAr.cxx
```diff
@@ -57,53 +57,33 @@ namespace larg4
...
         else
         {
-//            ThomasImel   = 0.099;
             DokeBirks[0] = 0.0003;
             DokeBirks[2] = 0.75;
         }
 
-        double Density = fDetProp->Density()/(CLHEP::g/CLHEP::cm3); // argon density at the temperature from Temperature()
+//        double Density = fDetProp->Density()/(CLHEP::g/CLHEP::cm3); // argon density at the temperature from Temperature()
+        double Density = LAr_Density;
```
absolution1, line 77: How wrong is the detector properties density value?
Is it worth looking at why the detector properties service is returning a non-useful value here?

larsim/IonizationScintillation/ISCalcNESTLAr.cxx
```diff
@@ -57,53 +57,33 @@ namespace larg4
...
-//        double delta    = 1*CLHEP::mm;
-//        double R0           = 1.568*CLHEP::um;    //Mozumder 1995
-
+        
         double DokeBirks[3];
-//        double ThomasImel = 0.00;
 
-        double eField = EFieldAtStep(fDetProp->Efield(), edep);
+//        double eField = EFieldAtStep(fDetProp->Efield(), edep);
+        double eField = EFieldAtStep(fDetProp->Efield(), edep) * 1000;  //conver kV to 1000 V
```
absolution1, line 64: Could a CLHEP unit conversion be used here (I think other CLHEP conversions are used in this file)?
e.g. https://www.zeuthen.desy.de/ILC/geant4/clhep-2.0.4.3/SystemOfUnits_8h-source.html

larsim/IonizationScintillation/ISCalcNESTLAr.h
```diff
@@ -30,6 +30,7 @@
 #define LAr_Z       (18)
 #define Density_LAr (1.393)
 #define LAr_W_Value (19.5*CLHEP::eV)
+#define LAr_Density (1.4)
```
absolution1, line 33: A fairly old school way of defining a value!
Could this be made as a private data member or something similar, rather than a preprocessor directive?

absolution1 (approved):
Looks fine to me.  I've left a couple of comments, which are obviously optional.
I'll grant L2 approval and let L1 provide their own comments :)

absolution1:
approve

lgarren:
@Healthborn  Dom has raised some important concerns.  We would like to see those concerns addressed  before approving this PR.

Healthborn:
1. How come you want to drop using the detector properties service here?
I drop using the service because the detector properties service gives the wrong value (4.5e19 vs 1.4).

2. Could a CLHEP unit conversion be used here (I think other CLHEP conversions are used in this file)?
Yes, it could.
I choose to implement this way, because of the facts:
 - the convention of in the original NEST Alg, where the efield is handle the in the way like: (ElectricField/1.0e3)
 - dividing it by 1000 with comment tells the code reviewers what the unit is, I think, more explicitly than using something like CLHEP unit  <-- that's where the original bug occurs (to convert some length unit using "CLHEP::cm")

3. A fairly old school way of defining a value! Could this be made as a private data member or something similar, rather than a preprocessor directive?
Yes, it could.
I have learned this from textbook and practised this way in a big commercial company, where they like it and I like it too.

absolution1:
Hi @Healthborn , thanks for responding.  I'll throw my two pennies in as I asked the original questions.

For Q1) it sounds like the detector properties service might be outputting an uninitialised value for the density?  I was thinking that it might be good to figure out why the detector properties service is not returning the correct value, rather than hardcoding what the value should be.

Q2) Relying on a comment can sometimes be tricky as comments can easily become stale if someone updates the code but does not update the comment.  I think unit conversions, like `CLHEP` are usually pretty clear, provided they are used properly.
A separate point about the conversion.  It looks like the kV to V conversion on line 64: https://github.com/LArSoft/larsim/blob/3a948d89cd662367745ebef95006fb9ff55a7aa4/larsim/IonizationScintillation/ISCalcNESTLAr.cxx#L64
is almost immediately cancelled out on line 67: https://github.com/LArSoft/larsim/blob/3a948d89cd662367745ebef95006fb9ff55a7aa4/larsim/IonizationScintillation/ISCalcNESTLAr.cxx#L67
Even if this was the original alg structure, it's a little bit confusing.  I also noticed that `eField` is used once more on line 87:
https://github.com/LArSoft/larsim/blob/3a948d89cd662367745ebef95006fb9ff55a7aa4/larsim/IonizationScintillation/ISCalcNESTLAr.cxx#L87
Are the units for `eField` still correct here?

3)  Sure, I think defining variables using preprocessor directives is mostly a style thing, even if it is more from the C-era.  I guess one downside is that `LAr_Density` might be exposed to whatever header/imp files `#include "ISCalcNESTLAr.h"`, which you probably don't want?

knoepfel:
> 3. A fairly old school way of defining a value! Could this be made as a private data member or something similar, rather than a preprocessor directive?
>  >  Yes, it could.
>  >  I have learned this from textbook and practised this way in a big commercial company, where they like it and I like it too.

Yes, preprocessing macros have been around for decades as the solution to problems that are now solved in far better ways with modern C++.  A few reasons against favoring macros in C++17 code:

- Macros are string-replacement facilities, the validity of which depends on where the macro is expanded.  This is in stark contrast to using C++ `constexpr` variables or functions.
- Macros do not respect scope.  A macro defined in one class, namespace, or file is accessible to any downstream scope unless the `#undef MY_MACRO` preprocessor command has been used--and there is no spell-checking to make sure you specified the correct macro name.

For this PR, the following would be much better:
```.cpp
namespace my_namespace {
  inline constexpr int LAr_Z{18};
  inline constexpr double Density_LAr{1.393};
  inline constexpr double LAr_W_Value{19.5 * CLHEP::eV};
  inline constexpr double LAr_Density{1.4};
}
```
It specifies the desired types (unlike a macro),  its scope is limited to the `my_namespace` scope (unlike a macro), and it will be evaluated at compile-time (like a macro).

aihimmel:
I would also put in a vote for understanding the values which come from the LArProperties service (maybe it's another units problem?) rather than hard-coding values in the algorithm. It allows the values to be changed by fhicl parameter, and since the values set in LArPropertiesService are part of the configuration stored in the file, someone downstream could be easily misled by what they find there if some other value was actually used in the algorithm.

lgarren:
@Healthborn  Are you able to address Alex Himmels concerns.

Healthborn:
The legacy NEST algorithm retrieves the density from **G4Material** at the step. I will check the refactored LArG4 to see whether the simulation retrieves the material density from **LArProperties service**. If it is, I agree that we'd better use the value from **LArProperties service** and troubleshoot why the service gives dramatic value.

lgarren:
code checks

lgarren:
@Healthborn larsim has been updated for geant4 10.6.  You will need to revisit some of your changes as they no longer merge cleanly with the head of develop.

lgarren:
code checks 

testing the code checks

lgarren:
code-checks

lgarren:
code-checks

testing the mechanism again

knoepfel:
@Healthborn, any update on this PR?

knoepfel:
Note that you may need to accommodate changes from LArSoft v09.

Healthborn:
> @Healthborn, any update on this PR?

Would you let me know what I am requested to do?

lgarren:
@Healthborn  Did you address the concerns about LArProperties?  Also, this PR fell out of sync with the head of develop before the move to larsoft v09.  There are merge conflicts.  That means that your changes overlap changes from someone else.  Would you please revisit your changes and make sure they are added such that the existing updates remain in place?

Healthborn:
> @Healthborn Did you address the concerns about LArProperties? Also, this PR fell out of sync with the head of develop before the move to larsoft v09. There are merge conflicts. That means that your changes overlap changes from someone else. Would you please revisit your changes and make sure they are added such that the existing updates remain in place?

What I was fixing is to correct the wrong value from the LArProperties. If the LArProperties gives the right value, it is not problem to use it in the future.

For the conflict, do you mean the files, which I was changing, was not up-to-date? Shall I checkout the up-to-date version on Jul 22 and then implement/re-commit my change or shall I do anything else?

lgarren:
@Healthborn  You will need to get the head of develop and look at the differences in the files you updated.

Healthborn:
> @Healthborn You will need to get the head of develop and look at the differences in the files you updated.

I'm wondering how I can get the head of develop. And if there is differences in the files I updated, what shall I do?

lgarren:
@Healthborn  please refer to https://cdcvs.fnal.gov/redmine/projects/larsoft/wiki/Working_with_GitHub

aihimmel:
@Healthborn -- I thought the status was that you were planning to switch to the LArProperties values, but that some work was needed to ensure that they were "sensible" for the NEST alogrithm (which I suspect is mostly a question of units). Have you checked that if you use the LArProperties values that NEST gives sensible results?

Healthborn:
> @Healthborn -- I thought the status was that you were planning to switch to the LArProperties values, but that some work was needed to ensure that they were "sensible" for the NEST alogrithm (which I suspect is mostly a question of units). Have you checked that if you use the LArProperties values that NEST gives sensible results?

In the original implementation of NEST algorithm, the value of material density is got from **G4Material**, since the (original) algorithm considers the several detector media: _argon_ or _xenon_. Therefore, when I migrate the original NEST algorithm for the refactored LArG4, I try to get the material density from **DetectorPropertiesService**. However, the value seems not sensible (**4.5e19** vs. **1.4**), which is not a question of unit but more likely a _parameter initialization_ problem.

To get value from either **DetectorPropertiesService** or **LArPropertiesServices** might have the same parameter initialization problem - if users does not assign a value to the density of liquid argon, the algorithm might give uncontrollable output.

aihimmel:
Hmmm.... strange. Perhaps it's an issue with the temperature which is set? It should be at 91 K by default. The density itself is not a fhicl parameter, but instead is a calculation based on the temperature. As you say, it's g/cm^3 so O(10^19) is not reasonable.

``` 218       /**
  219        * @brief Returns argon density at a given temperature
  220        * @param temperature the temperature in kelvin
  221        * @return argon density in g/cm^3
  222        *
  223        * Density is nearly a linear function of temperature.
  224        * See the NIST tables for details
  225        * Slope is between -6.2 and -6.1, intercept is 1928 kg/m^3.
  226        * This parameterization will be good to better than 0.5%.
  227        */
  228       virtual double Density(double temperature) const override;                          ///< g/cm^3
  229 
  230       // need to provide a definition, since the override above hides the inherited one
  231       virtual double Density() const override { return Density(Temperature()); }
  232 
  233       /// In kelvin.
  234       virtual double Temperature()                   const override { return fTemperature; }
```

larsim/IonizationScintillation/ISCalcNESTLAr.cxx
PetrilloAtWork: I have just modified the `DetectorPropertiesStandard` unit test to make it print the LAr density (`lardataalg` pull request lardataalg#14).
`DetectorPropertiesStandardTest standard_g4_icarus.fcl` prints a reasonable 1.39 kg/dm^3^.
So `DetectorProperties` _shouldn't_ be the problem. Passing the service provider through NESTed layers and resetting it on each event might be more of a challenge.

aihimmel: This is just more mysterious. I would have assumed the difficult part per @PetrilloAtWork's message would be just _getting_ a service handle. But, once we've gotten the handle, why would it get a nonsense value of the density? Are there instructions for running the unit tests? Perhaps @Healthborn can try running those, and confirm he's seeing different values for the density in the two contexts. It might help track down the issue.

PetrilloAtWork: The unit test is _art_-independent, and it utilises `DetectorPropertiesStandard` and `LArPropertiesStandard` (and `DetectorClocksStandard` and maybe `Geometry` with standard channel mapping: neither of these two dependencies should matter for the density). If DUNE uses these standard implementations, then the unit test is useful and can be run with `DetectorPropertiesStandardTest standard_g4_dune.fcl` (where the latter is any job configuration loading the relevant `DetectorProperties` configuration). The unit test should be in the `PATH` when `lardataalg` is set up or compiled.
If DUNE is using a custom implementation, the unit test won't exercise it. But I think David Adams also had some way to load services off-_art_, provided that it survived LArSoft `v09`.

larsim/PhotonPropagation/PDFastSimPAR_module.cc
```diff
@@ -48,13 +47,12 @@
 #include "larsim/PhotonPropagation/PhotonVisibilityService.h"
 #include "larsim/PhotonPropagation/PhotonVisibilityTypes.h" // phot::MappedT0s_t
 #include "larsim/PhotonPropagation/ScintTimeTools/ScintTime.h"
-
+#include "larsim/Simulation/LArG4Parameters.h"
```
PetrilloAtWork, line 50: Isn't this duplicating the next line?

Healthborn:
According to the discussion, I will close this PR and update the code with the latest code base.

PR #34, opened 2020-08-27
Update the arapuca dims every time, both for direct and reflected hits

ikatza:
On SBND new geometry we're changing the sizes of XARAPUCAS such that they are not longer the same as those of the ARAPUCAS.

So now every time the dimensions need to be updated such that they reflect the actual size of the current optical detector.

I'll prepare an equivalent commit for the equivalent code that uses new geant4 on another branch and PR.

The PR on sbndcode that will need this is: https://github.com/SBNSoftware/sbndcode/pull/6

### DIFF
--- a/larsim/LegacyLArG4/OpFastScintillation.cxx
+++ b/larsim/LegacyLArG4/OpFastScintillation.cxx
@@ -1493,11 +1493,9 @@ namespace larg4 {
     for (size_t const OpDet : util::counter(NOpChannels)) {
       if (!isOpDetInSameTPC(ScintPoint, fOpDetCenter.at(OpDet))) continue;
 
-      fydimension = fOpDetHeight.at(OpDet);
-      fzdimension = fOpDetLength.at(OpDet);
       // set detector struct for solid angle function
-      detPoint.h = fydimension;
-      detPoint.w = fzdimension;
+      detPoint.h = fOpDetHeight.at(OpDet);
+      detPoint.w =  fOpDetLength.at(OpDet);
       int const DetThis = VUVHits(Num, ScintPoint, fOpDetCenter[OpDet], fOpDetType[OpDet]);
       if (DetThis > 0) {
         DetectedNum[OpDet] = DetThis;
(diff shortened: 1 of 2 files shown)


larsim/LegacyLArG4/OpFastScintillation.cxx
```diff
@@ -1550,6 +1548,10 @@ namespace larg4 {
     for (size_t const OpDet : util::counter(NOpChannels)) {
       if (!isOpDetInSameTPC(ScintPoint, fOpDetCenter.at(OpDet))) continue;
 
+      // set detector struct for solid angle function
+      detPoint.h = fOpDetHeight.at(OpDet);
+      detPoint.w =  fOpDetLength.at(OpDet);
```
knoepfel: This seems to be an appropriate change.  However, instead of `detPoint` being a class data member, I would move it to be a stack variable that is passed in to the relevant functions (e.g.):
```.cpp
dims const det_point{fOpDetHeight.at(OpDet), fOpDetLength.at(OpDet)};
VUVHits(Num, ScintPoint, fOpDetCenter[OpDet], fOpDetType[OpDet], det_point);
```
I understand this changes the internal interface a little bit, but I think it is worth it.

ikatza: I'll take your advice on this, I'll make the changes in a bit.

lgarren:
@ikatza  I think we are waiting for changes from you.  Can you get them in soon?

knoepfel:
Well done, @ikatza.  No objections from L1.  Will refrain from approving until L2's give their consent.

ikatza:
This last push addresses Kyle suggestion, took the opportunity to make some more changes along those lines.

Commit 448dfce has nothing to do with ARAPUCAS, but it's too small of a change to call for another PR. It's an improvement in the management of the `CLHEP::RandGeneral`. There was a `new` not paired with any `delete`.

lgarren:
trigger build

lgarren:
CI tests are normal

ghost:
approve

lgarren:
approve

PR #35, opened 2020-08-28
Fix to LightSource module

PetrilloAtWork:
The `LightSource` module generates photons at specific points of the world.
The points can be specified by a text file, and each line of the text file will determine a volume and generate photons in there for the event. The next event will read the parameters from the next line of the text file, and so on.
When the text file is over, if further events are requested the module is supposed to start over with the settings at the beginning of the file, and then move on from there.

The wrap action at the end of the file was buggy so that the generation point for the first event after wrapping would be taken from uninitialised memory, while the other parameters would be unchanged. The next event would then correctly read the _first_ line of the steering file.

The bug has been fixed.

I have not investigated when this bug was introduced. My suspicion is that before my last changes, the bug was there but would effectively use the entire set of parameters from the previous event (no uninitialised memory access), hiding the issue and, in case of a single line in the steering file, producing the expected result.

Note: the branch is still based on `v08_62_01`. If this causes merge issues, please let me know and I'll try a port.


lgarren:
@PetrilloAtWork  Indeed, this PR contains merge conflicts.  Please rework it carefully for larsoft v09.

PetrilloAtWork:
It turns out the conflicts are caused by code alignment changes confusing GIT.
I'll provide a fixed branch soon.

PetrilloAtWork:
I believe I have fixed the conflicts.
Unfortunately it's quite hard to test them, since it appears like `v09_00_01` has _no full `Optical` GEANT4 simulation any more_.

PetrilloAtWork:
Incidentally, the bug I mention here has been now reported as [Redmine ticket #24890](https://cdcvs.fnal.gov/redmine/issues/24890).

larsim/EventGenerator/LightSource_module.cc
```diff
@@ -612,6 +614,22 @@ namespace evgen {
...
+  bool LightSource::readParametersFromInputFile() {
+    double x, y, z;
+    fInputFile >> x >> y >> z >> fT
+      >> fSigmaX >> fSigmaY >> fSigmaZ >> fSigmaT
+      >> fP >> fSigmaP >> fN;
+    fCenter = { x, y, z };
+    if (!fInputFile.good()) return false;
+
+    std::string dummy;
+    std::getline(fInputFile, dummy); // this can fail for what I care
```
knoepfel, line 628: What is the extra getline call for?

PetrilloAtWork: I am guessing here...
The format of the file being read includes one "parameter set" per line.
I imagine that the purpose was to discard any extra text (maybe comments?) after the parameter values.
The next call would then find the file ready at the beginning of the next line.

knoepfel: Well, maybe someday we can clean up this us.  But this infelicity isn't worth holding up approval.

lgarren:
trigger build

lgarren:
CI tests are normal

ghost:
approve

lgarren:
approve

PR #36, opened 2020-08-29
add a hit time RMS for IDE finding

wenqiang-gu:
When searching for IDE from a Hit, the default time start is defined as `hit->PeakTimeMinusRMS()` in MCCheater::BackTracker. However, both the longitudinal and transverse diffusion will broaden the hit width in a more realistic simulation. As a result, 1-RMS is not enough to cover all charges of an IDE. I suggest that we can change this time window in the configuration. The default value is still 1-RMS, while we suggest 3.5-RMS when it's necessary.

To chnge the default value, please add this line to your fhicl: 
`services.BackTrackerService.BackTracker.HitTimeRMS: 	      3.5`


yangtj207:
trigger build

yangtj207:
trigger build

yangtj207:
+1

knoepfel:
approve

PR #41, opened 2020-09-14
Performance improvements and back ported code to PDFastSimPAR

ikatza:
This PR ports the relevant code that has been developed, implemented and tested in OpFastScintillation on the LegacyLArG4 folder. Work done by me and @PetrilloAtWork.

This module still doesn't support reflected light, and I haven't added that myself. However, judging for the way it's currently implemented I'm guessing it will still use the same functions. Hence, I've taken the liberty to copy the functions used for reflected light in here. If requested I'm happy to implement the reflected light myself.

I've added some of the features to accommodate ICARUS-like geometry, but still some extra effort needs to go to make this module useful for ICARUS.

The formatting is all wonky. I used `clang-format` but got a different formatting. As a comment, I would like to mention that I don't like the new formatting at all.


aihimmel:
I think it's going to be quite hard to review this for accuracy given that basically every line in the file has changed. Do you have any validation studies you can point us to?

knoepfel:
@ikatza, yes, unfortunately the whitespace changes are too significant to review the meaningful differences here.  Can you please:

- Download the attached file and save it as `.clang-format`
[clang-format.txt](https://github.com/LArSoft/larsim/files/5220876/clang-format.txt)

- Run `clang-format -i -style=file <filename>` in the same directory that contains the `.clang-format` file
- Push the changes to this PR.

This will enable us to perform a review.  Also, stay tuned for an email discussing code format.

ikatza:
Thanks Kyle, it seems properly formatted now.

I have just managed to get this module working for SBND. You see, we're still transitioning to use this instead of the Legacy one. I'll try to get some validation tests done by next week. 

I'll do some for DUNE as well, but again will take me a while.

aihimmel:
Oh, I think I misunderstood. I thought this was related to work recently presented to the PD consortium which was introducing a change in the "physics" of the parameterization.

To confirm: this PR is to port the previous PR done on the legacy branch into the new LArG4? If so, then I'm much less concerned about the validation and scale of changes. We already went through those careful checks on the legacy part. I'd be happy to see whatever checks you have from the SBND side for now if that's OK, just to confirm that it is working and producing sane results.

ikatza:
Correct, I'm only bringing code from legacy onto here. There shouldn't be any change in physics, and hopefully everything should be transparent.

lgarren:
trigger build

lgarren:
@ikatza  There is a problem with the c7 build.   The failure involves unused variables.  See the log in the c7 CI report.

lgarren:
trigger build

lgarren:
trigger build

lgarren:
The CI tests are normal

lgarren:
@aihimmel do you want to approve this PR?

aihimmel:
I discussed with Iker on the side. He's working on some tests that the module is working as expected, and I think we should wait for those to be complete before merging. Since this isn't the default simulation yet, the CI tests wouldn't flag problems.

lgarren:
@ikatza @aihimmel  I am moving this back into "under discussion"  status so we know it is still in progress.  Please let us know when you are ready to have us look at this PR again.

ikatza:
I have run 1000 events. Using those events I ran `PDFastSim_PAR` with and without the changes in this PR. To make it easier to compare the runs I also made some temporary changes that shouldn't affect the consistency, and that I'm not committing. I list those down below.

Most events produce exactly the same amount of `detected fast photons` and `detected slow photons`. Summing photons across all the events I get 435 (1.7%) more fast photons, and 435 less (0.7%) slow photons. That is, the total number of detected photons is the same, but slightly different make up between fast and slow. I suspect this difference is related to random sampling. 

About the temporary changes. 

These ones below are experiment dependent and should be specified at a fcl file or somewhere else. I've set them to match SBND conditions, and to match the implicit conditions of the code without the changes in this PR.
```
  -  bool const fOnlyActiveVolume = false;
  +  bool const fOnlyActiveVolume = true;
    /// Allows running even if light on cryostats `C:1` and higher is not supported.
    /// Currently hard coded "no"
  -  bool const fOnlyOneCryostat = false;
  +  bool const fOnlyOneCryostat = true;
    /// Whether the cathodes are fully opaque; currently hard coded "no".
  - bool const fOpaqueCathode = false;
  + bool const fOpaqueCathode = true;
```

I also switch back to use TRandom
```
   - int hits_vuv = fRandPoissPhot->fire(GH_correction * hits_geo / cosine);    
   + double hits_rec = gRandom->Poisson(GH_correction * hits_geo / cosine);
   + int hits_vuv = std::round(hits_rec); // round to integer value, cannot have non-integer number of hits
```

Oddly enough I encountered an issue with the new geometry that I'm using, because PMTs are now specified as spheres, so had to go back to using this legacy idiom to categorise PMT and ARAPUCAS:
```
    - if (opDet.isBar()) {
    + if (strcmp(opDet.Shape()->IsA()->GetName(), "TGeoBBox") == 0) {
```
ultimately I believe `larcorealg/Geometry/OpDetGeo.h` should add a function to check if an optical detector is a sphere as we are improving the way we model PMTs to make them more realistic.

aihimmel:
Thanks for the careful checks. Approve.

knoepfel:
trigger build

lgarren:
The CI tests are normal.

lgarren:
approve

PR #43, opened 2020-09-23
Updates to semi-analytic light simulation model.

pgreen135:
Updates to semi-analytic light simulation model, presented in LArSoft Coordination meeting on 22-Sep-2020 (https://indico.fnal.gov/event/45687/).

Requires sbndcode and dunetpc feature branches with updated light simulation parameterisations to run:
sbndcode - https://github.com/SBNSoftware/sbndcode/pull/31
dunetpc - feature/pgreen_semianalytic_update


pgreen135:
I attempted to trigger the default_wf CI tests with the 3 feature branches (larsim, sbndcode, dunetpc) -  https://buildmaster.fnal.gov/buildmaster/job/lar_ci/9561. However, this does not appear to have worked properly - e.g. the dunetpc branch doesn't build. I'm guessing I have probably launched this incorrectly as I'm not familiar with the new method for the git repositories.

larsim/PhotonPropagation/PhotonVisibilityService_service.cc
```diff
@@ -733,64 +742,160 @@ namespace phot {
 
   //------------------------------------------------------
   void
-  PhotonVisibilityService::LoadTimingsForVUVPar(std::vector<double> v[9],
+  PhotonVisibilityService::LoadTimingsForVUVPar(std::vector<std::vector<double>> v[7],
```
knoepfel: I don't understand why the vector `v` is passed in by value.  Its entries are set by this function, but `v` is not used elsewhere in this function--so it appears to be unnecessary work being done.

knoepfel:
@pgreen135, which changes are essential to get in this week?  Is it just the bugfix or everything?

larsim/LegacyLArG4/OpFastScintillation.hh
```diff
@@ -252,11 +252,11 @@ namespace larg4 {
       std::vector<double> GetVisibleTimeOnlyCathode(double, int);*/
     // old timings -- to be deleted
 
-    void getVUVTimes(std::vector<double>& arrivalTimes, double distance_in_cm);
-    void generateParam(const size_t index);
+    void getVUVTimes(std::vector<double>& arrivalTimes, const double &distance_in_cm, const size_t &angle_bin);
+    void generateParam(const size_t &index, const size_t &angle_bin);
```
knoepfel: Any type whose size is roughly the size of a pointer should be passed in by value, and not by reference.  For each of the changes arguments above, the reference should be removed.

pgreen135:
> @pgreen135, which changes are essential to get in this week? Is it just the bugfix or everything?

Everything, this is to have the updated model and updated parameterisations available for SBND prior to the upcoming SBN production.

larsim/LegacyLArG4/OpFastScintillation.hh
```diff
@@ -373,35 +374,48 @@ namespace larg4 {
     double Rectangle_SolidAngle(Dims const&  o, const std::array<double, 3> v);
     // solid angle of circular aperture calculation functions
     double Disk_SolidAngle(const double d, const double h, const double b);
+    // solid angle of a dome aperture calculation functions
+    double Omega_Dome_Model(const double &distance, const double &theta) const;
```
knoepfel: Remove the references here--should pass by value.

larsim/LegacyLArG4/OpFastScintillation.hh
```diff
@@ -373,35 +374,48 @@ namespace larg4 {
...
+    // flat PDs
+    bool fIsFlatPDCorr;
+    std::vector<std::vector<double>> fGHvuvpars_flat;
+    std::vector<double> fborder_corr_angulo_flat;
+    std::vector<std::vector<double>> fborder_corr_flat;
+    // dome PDs
+    bool fIsDomePDCorr;
+    std::vector<std::vector<double>> fGHvuvpars_dome;
+    std::vector<double> fborder_corr_angulo_dome;
+    std::vector<std::vector<double>> fborder_corr_dome;
```
knoepfel, line 393: Unless you expect flat and dome PDs to be enabled at the same time, there is no reason to have separate data members for both--that just needlessly increases the size of the class and adds maintenance.  The only thing that appears essential is:
```.cpp
bool fIsFlatPDCorr;
bool fIsDomePDCorr;
std::vector<std::vector<double>>  fGHvuvpars;
std::vector<double> fborder_corr_angulo;
std::vector<std::vector<double>> fborder_corr;
```
Then, the way the vectors are filled just depends on the values of `fIsFlatPDCorr` and `fIsDomePDCorr`.

larsim/LegacyLArG4/OpFastScintillation.hh
```diff
@@ -373,35 +374,48 @@ namespace larg4 {
...
+    // correction parameters for VIS Nhits estimation
+    double fdelta_angulo_vis;
+    // flat PDs
+    std::vector<double> fvis_distances_x_flat;
+    std::vector<double> fvis_distances_r_flat;
+    std::vector<std::vector<std::vector<double>>> fvispars_flat;
+    // dome PDs
+    std::vector<double> fvis_distances_x_dome;
+    std::vector<double> fvis_distances_r_dome;
+    std::vector<std::vector<std::vector<double>>> fvispars_dome;
```
knoepfel, line 406: See comment above about duplicated vectors--they likely are not necessary here.

larsim/PhotonPropagation/PhotonVisibilityService.h
```diff
@@ -256,33 +295,47 @@ namespace phot {
...
+    //parametrization exists for DUNE-SP and for SBND
+    double fdelta_angulo_vuv;
+    // flat PDs
+    std::vector<std::vector<double>> fGHvuvpars_flat;
+    std::vector<double> fborder_corr_angulo_flat;
+    std::vector<std::vector<double>> fborder_corr_flat;
+    // dome PDs
+    std::vector<std::vector<double>> fGHvuvpars_dome;
+    std::vector<double> fborder_corr_angulo_dome;
+    std::vector<std::vector<double>> fborder_corr_dome;
```
knoepfel, line 324: See comment above about duplicated vectors--they are probably not necessary here.

larsim/PhotonPropagation/PhotonVisibilityService.h
```diff
@@ -256,33 +295,47 @@ namespace phot {
...
-    std::vector<std::vector<std::vector<double>>> fVIS_BORDER_correction;
+    double fdelta_angulo_vis;
+    // flat PDs
+    std::vector<double> fvis_distances_x_flat;
+    std::vector<double> fvis_distances_r_flat;
+    std::vector<std::vector<std::vector<double>>> fvispars_flat;
+    // dome PDs
+    std::vector<double> fvis_distances_x_dome;
+    std::vector<double> fvis_distances_r_dome;
+    std::vector<std::vector<std::vector<double>>> fvispars_dome;
```
knoepfel, line 335: Again, duplicated vectors based on 'flat' vs. 'dome' are likely not necessary.

larsim/PhotonPropagation/PhotonVisibilityService_service.cc
```diff
@@ -313,57 +313,66 @@ namespace phot {
...
     }
 
     if (fUseNhitsModel) {
       std::cout << "Loading semi-analytic mode models" << std::endl;
       // VUV
-      fGH_PARS = p.get<std::vector<std::vector<double>>>("GH_PARS");
-      std::vector<double> v0(2, 0.0);
-      fBORDER_correction = p.get<std::vector<double>>("BORDER_correction", v0);
+      fIsFlatPDCorr = p.get<bool>("FlatPDCorr", false);
+      fIsDomePDCorr = p.get<bool>("DomePDCorr", false);
```
knoepfel, line 348: Are these supposed to be disjoint options?  What happens if a user specifies both as `false` or both as `true`?

larsim/PhotonPropagation/PhotonVisibilityService_service.cc
```diff
@@ -733,64 +742,160 @@ namespace phot {
...
+    vis_distances_r_dome = fvis_distances_r_dome;
+    vispars_dome = fvispars_dome;
   }
 
+
+  // placeholder functions for loading old defunct sets of parameterisations
+  // required to allow (not yet in use) new LArG4 semi-analytic model to compile
+  // calls to these functions to be replaced in new LArG4 once updated to new method, and placeholders removed
+  // timings
+  void PhotonVisibilityService::LoadTimingsForVUVPar(std::vector<double> v[9],
```
knoepfel: Another example where `v` appears to be passed by value,  set,  and then discarded.

larsim/PhotonPropagation/PhotonVisibilityService_service.cc
```diff
@@ -733,64 +742,160 @@ namespace phot {
...
+    border_corr_flat = fborder_corr_flat;
+  }
+  void PhotonVisibilityService::LoadGHDome( std::vector<std::vector<double>>  &GHvuvpars_dome,
+                   std::vector<double> &border_corr_angulo_dome,
+                   std::vector<std::vector<double>> &border_corr_dome) const
+  {
+    GHvuvpars_dome = fGHvuvpars_dome;
+    border_corr_angulo_dome = fborder_corr_angulo_dome;
+    border_corr_dome = fborder_corr_dome;
+  }
```
knoepfel: At the very least, you should have a guard in these functions in case `isFlatPDCorr` or `isDomePDCorr` is false:
```.cpp
void LoadGHFlat(...) {
  if (not fIsFlatPDCorr) return;
  ...
}
```
Also, if the 'flat' and 'dome' options are disjoint, then there is no need to have both functions.

larsim/PhotonPropagation/PhotonVisibilityService_service.cc
```diff
@@ -733,64 +742,160 @@ namespace phot {
...
+    vispars_flat = fvispars_flat;
+  }
+  void PhotonVisibilityService::LoadVisParsDome(std::vector<double> &vis_distances_x_dome,
+                       std::vector<double> &vis_distances_r_dome,
+                       std::vector<std::vector<std::vector<double>>> &vispars_dome) const
+  {
+    vis_distances_x_dome = fvis_distances_x_dome;
+    vis_distances_r_dome = fvis_distances_r_dome;
+    vispars_dome = fvispars_dome;
   }
```
knoepfel: See above regarding guarding against disabled 'flat' or 'dome' corrections, and whether it's necessary to have both functions.

larsim/LegacyLArG4/OpFastScintillation.cxx
```diff
@@ -274,29 +279,38 @@ namespace larg4 {
...
-        const size_t num_params =
-          (fmax_d - 25) /
-          fstep_size; // for d < 25cm, no parameterisaton, a delta function is used instead // TODO: unhardcode
+        const size_t num_params = (fmax_d - fmin_d) / fstep_size; // for d < fmin_d, no parameterisaton, a delta function is used instead
+        size_t num_angles = std::round(90/fangle_bin_timing_vuv);
         std::vector<TF1> VUV_timing_temp(num_params, TF1());
-        VUV_timing = VUV_timing_temp;
+        for (size_t i = 0; i < num_angles; ++i) {
+          VUV_timing.push_back(VUV_timing_temp);
+        }
```
knoepfel: This can be abbreviated:
```.cpp
VUV_timing = std::vector(num_angles, std::vector(num_params, TF1{}));
```
I encourage you to follow a similar pattern for setting the vector member variables.

knoepfel (commented):
Thanks for the PR, @pgreen135.  There are a number of comments for your consideration.  Probably the most important is whether the `fIsDomeCorr` and `fIsFlatCorr` variables can both be `true`.  If yes, then some of my comments are unnecessary.  If, however, they cannot both be `true` at the same time, then there is some code duplication that should be removed.

Thanks, and happy to continue the discussion.

pgreen135:
> Thanks for the PR, @pgreen135. There are a number of comments for your consideration. Probably the most important is whether the `fIsDomeCorr` and `fIsFlatCorr` variables can both be `true`. If yes, then some of my comments are unnecessary. If, however, they cannot both be `true` at the same time, then there is some code duplication that should be removed.
> 
> Thanks, and happy to continue the discussion.

Hi Kyle, both `fIsDomeCorr` and `fIsFlatCorr` can be true at the same time, the idea is to allow the appropriate parameters to be used depending on the OpDet shape -- in the case of SBND there is a mixture of PMTs ("dome" type) and Arapucas ("flat" type). In the case they are both false then the model won't work because no parameters have been provided - I could add an error message related to that?

I agree with all the other comments and I will address these now and push an update, thanks.

knoepfel:
Okay, thanks for the clarification Patrick.  Feel free, then, to resolve all of the conversations related to duplicate code.  If 'dome' and 'flat' type should not both be false, then the code should throw an exception with a helpful message indicating the configuration error.

pgreen135:
@knoepfel I have updated the code addressing I believe all of the above comments

knoepfel:
trigger build

lgarren:
trigger build #43 SBNSoftware/sbndcode#31 dunetpc@feature/pgreen_semianalytic_update

lgarren:
c7 CI is OK: https://dbweb5.fnal.gov:8443/LarCI/app/view_builds/index?offset=0&builds=lar_ci/9571&builds=
e19 CI in progress:  https://dbweb5.fnal.gov:8443/LarCI/app/view_builds/index?offset=0&builds=lar_ci/9572&builds=

pgreen135:
I have just pushed an update to the generic transport time parameters to make them compatible with the model changes. This was the reason the icaruscode CI tests failed previously (I had not realised this was in use by them). The sbndcode and dunetpc code failed due to missing the new feature branches with the updated specific parameters for those detectors.

lgarren:
I have just triggered the CI builds by hand (again).  Patrick, the command I used is
`trigger --build-delay 0 --cert /tmp/x509up_u1147 --version develop  --workflow default_wf --revisions "LArSoft/larsim#43 SBNSoftware/sbndcode#31 dunetpc@feature/pgreen_semianalytic_update" --quals e19:prof --force-platform slf7`

lgarren:
The CI tests are normal
https://dbweb5.fnal.gov:8443/LarCI/app/view_builds/index?offset=0&builds=lar_ci/9573&builds=
https://dbweb0.fnal.gov/LarCI/app/view_builds/index?offset=0&builds=lar_ci/9574&builds=

lgarren:
approve

(Will merge when building the release since it needs accompanying experiment feature branches)

pgreen135:
@lgarren I went over the results from the CI tests carefully this morning and noticed there was still an issue remaining for the generic transport time parameters used by icarus that was causing errors when extrapolating one of the parameters at very large distances. The push above fixes this issue (just a very minor change to one of the parameters). I have re-run the full set of CI tests  - see lar_ci/9585 - and these now all look as expected including icarus. This push has reverted the approval from yesterday so this will need to be repeated, sorry for being a nuisance!

lgarren:
The release was tagged and built last night.  I will need to make a new release.

lgarren:
approve

PR #44, opened 2020-10-10
To access OpFastScintillation class from sbncode

marcodeltutto:
SBN needs to access `OpFastScintillation::detectedDirectHits()` and `OpFastScintillation::detectedReflecHits()` in order to get an estimation of the photons that arrive to optical detectors using the semi-analytical approach instead of the photon library. Currently, those two methods are private. This PR changes them to public. 

`detectedDirectHits` and `detectedReflecHits` should be provided by a service (similarly to the `PhotonVisibilityService`) and this is planned to be done in the future when things will be moved to `larg4`. For now, we need to access them in this way.

### DIFF
--- a/larsim/LegacyLArG4/OpFastScintillation.cxx
+++ b/larsim/LegacyLArG4/OpFastScintillation.cxx
@@ -2192,15 +2192,15 @@ namespace larg4 {
 
   // ---------------------------------------------------------------------------
 
-  constexpr double
-  acos_table(const double x)
-  {
-    if (x < 0. || x > 1.) {
-      std::cout << "Range out of bounds in acos_table, only defined in [0, 1]" << std::endl;
-      exit(0);
-    }
-    return acos_arr[std::round(acos_bins * x)];
-  }
+  // constexpr double
+  // acos_table(const double x)
+  // {
+  //   if (x < 0. || x > 1.) {
+  //     std::cout << "Range out of bounds in acos_table, only defined in [0, 1]" << std::endl;
+  //     exit(0);
+  //   }
+  //   return acos_arr[std::round(acos_bins * x)];
+  // }
 
   double
   fast_acos(double x)
--- a/larsim/LegacyLArG4/OpFastScintillation.cxx
+++ b/larsim/LegacyLArG4/OpFastScintillation.cxx
@@ -581,7 +581,6 @@ namespace larg4 {
 
     // Get the visibility vector for this point
     assert(fPVS);
-    NOpChannels = fPVS->NOpChannels();
 
     G4int nscnt = 1;
     if (Fast_Intensity && Slow_Intensity) nscnt = 2;
@@ -740,7 +739,7 @@ namespace larg4 {
       //std::cout << "++++++++++++" << Num << "++++++++++" << std::endl;
 
       // here we go: now if visibilities are invalid, we are in trouble
-      //if (!Visibilities && (NOpChannels > 0)) {
+      //if (!Visibilities && (fPVS->NOpChannels() > 0)) {
       //  throw cet::exception("OpFastScintillator")
       //    << "Photon library does not cover point " << ScintPoint << " cm.\n";
       //}
@@ -750,7 +749,7 @@ namespace larg4 {
       // detected photons from direct light
       std::map<size_t, int> DetectedNum;
       if (Visibilities && !usesSemiAnalyticModel()) {
-        for (size_t const OpDet : util::counter(NOpChannels)) {
+        for (size_t const OpDet : util::counter(fPVS->NOpChannels())) {
           if (fOpaqueCathode && !isOpDetInSameTPC(ScintPoint, fOpDetCenter.at(OpDet))) continue;
           int const DetThis = std::round(G4Poisson(Visibilities[OpDet] * Num));
           if (DetThis > 0) DetectedNum[OpDet] = DetThis;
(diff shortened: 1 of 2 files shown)


absolution1:
trigger build

larsim/LegacyLArG4/OpFastScintillation.hh
```diff
@@ -464,8 +464,8 @@ namespace larg4 {
   double model_far(double*, double*);
 
   static const size_t acos_bins = 2000000;
-  static std::array<double, acos_bins + 1>
-    acos_arr; // to get minimum resolution of 0.0000005 in [0,1]
+  // static std::array<double, acos_bins + 1>
+  //   acos_arr; // to get minimum resolution of 0.0000005 in [0,1]
```
knoepfel: Recommend removing the lines instead of commenting them out.

larsim/LegacyLArG4/OpFastScintillation.cxx
```diff
@@ -2192,15 +2192,15 @@ namespace larg4 {
...
-  }
+  // constexpr double
+  // acos_table(const double x)
+  // {
+  //   if (x < 0. || x > 1.) {
+  //     std::cout << "Range out of bounds in acos_table, only defined in [0, 1]" << std::endl;
+  //     exit(0);
+  //   }
+  //   return acos_arr[std::round(acos_bins * x)];
+  // }
```
knoepfel: Recommend removing instead of commenting out.

larsim/LegacyLArG4/OpFastScintillation.hh
```diff
@@ -259,6 +259,13 @@ namespace larg4 {
     void getVISTimes(std::vector<double>& arrivalTimes, const TVector3 &ScintPoint, const TVector3 &OpDetPoint);
     // Visible component timing parameterisation
 
+    void detectedDirectHits(std::map<size_t, int>& DetectedNum,
+                            const double Num,
+                            geo::Point_t const& ScintPoint);
+    void detectedReflecHits(std::map<size_t, int>& ReflDetectedNum,
+                            const double Num,
+                            geo::Point_t const& ScintPoint);
```
knoepfel: Now that these functions are public, is it possible to mark them as `const`?  This may require marking other member functions as `const` also.  The goal is to avoid adjusting the internal state of the class from outside.

larsim/LegacyLArG4/OpFastScintillation.cxx
aihimmel: Removing this was general cleanup? Not that it's a bad idea, just trying to understand since it seems unrelated to the main purpose of the PR.

larsim/LegacyLArG4/OpFastScintillation.cxx
```diff
@@ -1577,7 +1577,7 @@ namespace larg4 {
 
     // detemine hits on each PD
     const std::array<double, 3> hotspot = {plane_depth, ScintPoint.Y(), ScintPoint.Z()};
-    for (size_t const OpDet : util::counter(NOpChannels)) {
+    for (size_t const OpDet : util::counter(fPVS->NOpChannels())) {
```
aihimmel, line 1579: Should the NOpChannels member be removed? Where else is it being used? If not, why not keep it here?

larsim/LegacyLArG4/OpFastScintillation.cxx
```diff
@@ -1501,7 +1501,7 @@ namespace larg4 {
                                           const double Num,
                                           geo::Point_t const& ScintPoint)
   {
-    for (size_t const OpDet : util::counter(NOpChannels)) {
+    for (size_t const OpDet : util::counter(fPVS->NOpChannels())) {
```
aihimmel, line 1503: Should the NOpChannels member be removed? Where else is it being used? If not, why not keep it here?

larsim/LegacyLArG4/OpFastScintillation.hh
marcodeltutto: Yes, good point, thanks. Done with commit 8721ec9.

larsim/LegacyLArG4/OpFastScintillation.hh
marcodeltutto: Done with d829239.

larsim/LegacyLArG4/OpFastScintillation.cxx
marcodeltutto: Lines removed with d829239.
It was cleanup but it was also related to this PR. The variable `acos_arr` was unused, causing `sbndcode` to fail compilation. I can of course change to compiler flags in `sbndcode` when it compiles this part, but I think it is cleaner to just remove the unused variable.

larsim/LegacyLArG4/OpFastScintillation.cxx
marcodeltutto: Yes, thanks, done with a4c7c2e.

larsim/LegacyLArG4/OpFastScintillation.cxx
marcodeltutto: Yes, thanks, done with a4c7c2e.

larsim/LegacyLArG4/OpFastScintillation.cxx
marcodeltutto: It removed the `NOpChannels` member because it was not initialized in the constructor, and so depending on how you call this class you may not get `NOpChannels` properly initialized. I can either move the initialization to the constructor, or just remove this variable as the number of op dets is provided by the service. I preferred to go the second way as I thought it was cleaner.

marcodeltutto:
Hello, if possible, we would need this in the next LArSoft release. Please let me know if anything else is needed from my side. @knoepfel @aihimmel

lgarren:
trigger build

larsim/LegacyLArG4/OpFastScintillation.cxx
aihimmel: OK, sounds good good.

aihimmel:
Looks OK to me. Sorry for the delay. Approve.

wesketchum:
Thanks @aihimmel.

Regarding the failed/warning on the tests, probably an expert eye needed. It's not clear to me that these are related to this PR, or to other underlying issues common to other tests. Sorry if I'm adding noise, but just trying to help push this along.

lgarren:
@wesketchum  If you can help get those CI tests updated, we will be delighted. ;)

lgarren:
approve

PR #46, opened 2020-10-26
added option to bypass ifdh copying of corsika shower db files and use a direct file path instead

pgreen135:
I've added an option to use a direct file path for a corsika database file instead of copying using ifdh, analogous to option in genie. I have been using this approach on Theta (ALCF) to a). minimise the copying required at large scale and b). to allow ifdh to be avoided, due to frequent unexplained segfaults when copying files with ifdh on Theta's filesystem (possibly a memory issue? the seg faults occur while copying db files or flux files, but sporadically and not always at the same point in the code). 

This is controlled by a new fhicl parameter, fShowerCopyType, which can either be "IFDH" (default) or "DIRECT". As default the standard IFDH copying is used, so unless this parameter is set this change should be invisible.


absolution1:
trigger build

lgarren:
trigger build

the uboone suite is now updated

absolution1:
approve

lgarren:
CI tests are normal

lgarren:
approve

PR #47, opened 2020-11-04
Fix misleading message in PhotonVisibilityService

PetrilloAtWork:
A message about photon library build mode appeared even in some cases when library build mode was not active.
This simple fix (to an issue that I suspect _I_ may have introduced long ago) prevents that.

Now, this is a very bad thing that I am doing, by creating a fix using GitHub editor (meaning, I did not even compile the proposed fix and I am relying on the C.I. test). I might be using too much Python...

Also: this change can be aggregated to others before being scheduled for merge, since it is not urgent.


aihimmel:
Change looks fine to me. I suspect this never got updated when we moved to PVS configurations which neither loaded a library nor were library build jobs. Thanks Gianluca.

Approve.

lgarren:
trigger build

knoepfel:
approve

PR #48, opened 2020-11-05
[FEAT] Update code to support Reflected Photons and SimPhotons

Healthborn:
In PDFastSimPVS module:
- the reflected photons and sim photons will be supported.
In PDFastSimPAR module
- the reflected photons and sim photons will be supported.
- the VISHits and related functions are updatged (ported from OpFastScintillation.cxx)

PR #14, opened 2020-04-16
Performance, memory and styling improvements in OpFastScintillation

ikatza:
Have done some major refactoring of the code inside `LegacyLArG4/OpFastScintillation`. 

Testing on a sbndgpvm on Cosmics + Neutrino events; the most cpu and memory hungry of events, I have achieved close to half of running time and a comparable memory footprint.

Despite using some fast approximations to the most used functions there is no measurable bias in the OpHits.

Along the way I have also address some of the underlying potential bugs that were hidden in the code. The most salient one had to with  a domain error exception calling the boost elliptic integral functions, as far as I can tell this would happen due to a floating point error bloating.

PR #50, opened 2020-11-12
Updating PDFastSimPAR with new semi-analytic simulation models previously implemented in legacy LArG4

pgreen135:
Updates to the refactored LArG4 parameterised light simulation module with the refined model(s) implemented in legacy LArG4 in PR #43. The changes to the models are detailed in https://indico.fnal.gov/event/45687/ (presented in Sep before the update to the legacy code).

I have also added support for SimPhotons and Reflected photons in PDFastSimPAR following the approach used by Muve (see PR #48) and used in legacy LArG4.


larsim/PhotonPropagation/PDFastSimPAR_module.cc
```diff
@@ -216,6 +203,18 @@ namespace {
...
+  template <typename TReal>
+  inline constexpr double
+  dist(const TReal* x, const TReal* y, const unsigned int dimension)
+  {
+    double d = 0.;
+    for (unsigned int p = 0; p < dimension; ++p) {
+      d += (*(x + p) - *(y + p)) * (*(x + p) - *(y + p));
+    }
+    return std::sqrt(d);
+  }
```
knoepfel: Several comments:

- `std::sqrt` is not a `constexpr` function, even though compilers like GCC will support it.  However, Clang will likely choke on this.  Best to remove the `constexpr` keyword on line 208.
- This function is only used for iterating through the elements of type `std::array<double, 3u>` and `TVector3`.  I encourage using one or the other types.  Then there is no need to have a function template.
- The implementation requires the user to provide pointers to the elements and also a limit value, which is used to access memory in an unchecked way (i.e. what happens if someone specifies a `dimension` of 4).

For these reasons, I encourage a different implementation:

```.cpp
#include "cetlib/pow.h"

double dist(std::array<double, 3u> const& a, 
            std::array<double, 3u> const& b,
            size_t start = 0)
{
  double d = 0.;
  for (; start != size(a); ++start) {
    d += cet::square(a[start] - b[start]);
  }
  return std::sqrt(d);
}
```
And then the use of it is:

```.cpp
double r = dist(ScintPoint_v, fcathode_centre, 1);
```
assuming both types are the same.

larsim/PhotonPropagation/PDFastSimPAR_module.cc
```diff
@@ -418,8 +445,30 @@ namespace phot {
     std::cout << "PDFastSimPAR Module Construct" << std::endl;
 
     Initialization();
-    produces<std::vector<sim::SimPhotonsLite>>("par");
-    produces<std::vector<sim::OpDetBacktrackerRecord>>("par");
+    art::ServiceHandle<sim::LArG4Parameters const> lgp;
+    if (lgp->UseLitePhotons())
+    {
+        std::cout << "Use Lite Photon." << std::endl;
```
knoepfel: Suggest using `mf::LogInfo(...)` instead of `std::cout`.

larsim/PhotonPropagation/PDFastSimPAR_module.cc
```diff
@@ -429,18 +478,32 @@ namespace phot {
...
-    std::unique_ptr<std::vector<sim::OpDetBacktrackerRecord>> opbtr(
-      new std::vector<sim::OpDetBacktrackerRecord>);
-    std::unique_ptr<std::vector<sim::SimPhotonsLite>> phlit(new std::vector<sim::SimPhotonsLite>);
+    nOpChannels = fPVS->NOpChannels();
 
-    auto& photonLiteCollection(*phlit);
-    photonLiteCollection.resize(nOpChannels);
-    for (unsigned int i = 0; i < nOpChannels; i++) {
-      photonLiteCollection[i].OpChannel = i;
+    std::unique_ptr< std::vector< sim::SimPhotons > >             phot   (new std::vector<sim::SimPhotons>);
```
knoepfel: Each of these lines can be written more concisely as:
```.cpp
auto phot = std::make_unique<std::vector<sim::SimPhotons>>();
```

larsim/PhotonPropagation/PDFastSimPAR_module.cc
```diff
@@ -465,51 +528,122 @@ namespace phot {
       double edeposit = edepi.Energy() / nphot;
       double pos[3] = {edepi.MidPointX(), edepi.MidPointY(), edepi.MidPointZ()};
       geo::Point_t const ScintPoint = {pos[0], pos[1], pos[2]};
+      G4ThreeVector ScintPoint_v = {pos[0], pos[1], pos[2]};
```
knoepfel: Can the interface of `propagationTime(...)` be adjusted so this extra variable is not necessary?

larsim/PhotonPropagation/PDFastSimPAR_module.cc
```diff
@@ -667,36 +818,105 @@ namespace phot {
...
+                                   const double NumSlow,
+                                   geo::Point_t const& ScintPoint)
+  {
+    for (size_t const OpDet : util::counter(nOpChannels)) {
+      if (!isOpDetInSameTPC(ScintPoint, fOpDetCenter.at(OpDet))) continue;
+
+      // set detector struct for solid angle function
+      const PDFastSimPAR::OpticalDetector op{
+        fOpDetHeight.at(OpDet), fOpDetLength.at(OpDet),
+        fOpDetCenter.at(OpDet), fOpDetType.at(OpDet)};
```
knoepfel: Is the `.at(OpDet)` necessary instead of `[OpDet]`?  `at` introduces additional bounds-checking that can be expensive when not necessary.

larsim/PhotonPropagation/PDFastSimPAR_module.cc
```diff
@@ -705,162 +925,296 @@ namespace phot {
...
+    // 1). calculate total number of hits of VUV photons on
+    // reflective foils via solid angle + Gaisser-Hillas
+    // corrections:
+
+    // set plane_depth for correct TPC:
+    double plane_depth;
+    if (ScintPoint.X() < 0.) { plane_depth = -fplane_depth; }
+    else {
+      plane_depth = fplane_depth;
+    }
```
knoepfel: Suggest more concise:
```.cpp
double const plane_depth = ScintPoint.X() < 0. ? -fplane_depth : fplane_depth;
```

larsim/PhotonPropagation/PDFastSimPAR_module.cc
```diff
@@ -705,162 +925,296 @@ namespace phot {
...
+    double distance_cathode = std::abs(plane_depth - ScintPoint.X());
+    // calculate hits on cathode plane via geometric acceptance
+    double cathode_hits_geo_fast = std::exp(-1. * distance_cathode / fL_abs_vuv) *
+                              (solid_angle_cathode / (4. * CLHEP::pi)) * NumFast;
+    double cathode_hits_geo_slow = std::exp(-1. * distance_cathode / fL_abs_vuv) *
+                              (solid_angle_cathode / (4. * CLHEP::pi)) * NumSlow;
+
+    // determine Gaisser-Hillas correction including border effects
+    // use flat correction
+    double r = std::sqrt(std::pow(ScintPoint.Y() - fcathode_centre[1], 2) + std::pow(ScintPoint.Z() - fcathode_centre[2], 2));
```
knoepfel: Suggest more concise:
```.cpp
double r = std::hypot(ScintPoint.Y() - fcathode_center[1], fScintPoint.Z() - fcathode_center[2]);
```

knoepfel (commented):
Thanks, @pgreen135, for the PR. Some comments below for your consideration.  There is a bit of code duplication in the module.  Some of my comments are geared toward removing such duplication, but a thorough review of the code would be required to come up with a trimmer design.  Happy to discuss further.

pgreen135:
> Thanks, @pgreen135, for the PR. Some comments below for your consideration. There is a bit of code duplication in the module. Some of my comments are geared toward removing such duplication, but a thorough review of the code would be required to come up with a trimmer design. Happy to discuss further.

Thanks for the comments @knoepfel, I will update the code with each of the suggested changes above. Regarding a review / restructuring of the module - this PR is purely to port the recent updates made in the legacy simulation into the refactored version. If I'm understanding your comment correctly, structural changes to this module is something that should be discussed with Mu Wei who wrote the framework - I have just tried to keep to his design in this PR (at his request).

aihimmel:
Right. I agree with @pgreen135 -- let's keep the scope limited to keeping legacy and new LArG4 in sync. I suggest we circle back around to more substantial cleanup once everyone has moved over to the new LArG4 and we don't need to support both systems. And frankly, I'm not sure this is even the biggest fish to fry on the cleanup front. (Looking at you, SimPhotonCounter_module.cc)

larsim/PhotonPropagation/PDFastSimPAR_module.cc
pgreen135: @knoepfel  this function & interface was added by Iker De Icaza in I believe PR #14 in the legacy version, and I copied it over from there as it is used in the legacy code. I agree it is fatally flawed and frankly I am unsure of its purpose over using standard methods. Not only that, but the geo:Point_t objects are converted to std::array<double, 3u> specifically to use this function, which doesn't seem to make any sense. I've removed all of this entirely, and swapped to just using std::hypot instead. Let me know if that is reasonable.

pgreen135:
I have pushed an update that implements all of the changes suggested above, with the exception of the new dist(...) function. Instead I've removed that function entirely - see my response to that particular comment.

lgarren:
trigger build

lgarren:
@pgreen135  There is a c7 build error.  I think just an unused variable.  
https://dbweb5.fnal.gov:8443/LarCI/app/storage/docs/2020/11/19/buildLArSoft.log

lgarren:
@pgreen135 Sorry - I see you already fixed it.  Thanks!

lgarren:
trigger build

lgarren:
approve

PR #51, opened 2020-11-13
larsim now depends on larg4

lgarren:
Larsim will now depend on larg4.  We have always intended to do this, and now is the time.


lgarren:
trigger build

lgarren:
approve

lgarren:
Because this changes the table file, the PR will be merged when tagging the next release.

PR #53, opened 2020-11-18
Volume name feature for RadioGen event generator.

PetrilloAtWork:
Features have been added to `RadioGen` event generator:
* volumes can be added by name instead of coordinates
* generation times can be specified nuclide by nuclide, or one for all (or there is a not-so-useful default)
* added Doxygen documentation

These and other features have been discussed long time ago with @tomjunk, but at this time I have to be realistic and accept I will not have the time to add them.
So I invite @tomjunk in particular to check this proposal, and I will fix anything related to the _currently proposed_ features.

Note that ICARUS is already using the new features (specifying `volCryostat` is way more convenient and maintainable than specifying in the configuration file the coordinates of two different volumes).

### DIFF
--- a/larsim/EventGenerator/RadioGen_module.cc
+++ b/larsim/EventGenerator/RadioGen_module.cc
@@ -897,12 +897,21 @@ namespace evgen{
     auto const& detInfo
       = *(lar::providerFrom<detinfo::DetectorPropertiesService>());
     
-    TPCelectronics_tick const startTick { -int(detInfo.ReadOutWindowSize()) };
-    TPCelectronics_tick const endTick { detInfo.NumberTimeSamples() };
+    //
+    // we take a number of (TPC electronics) ticks before the trigger time,
+    // and we go to a number of ticks after the trigger time;
+    // that shift is one readout window size
+    //
+    
+    auto const trigTimeTick
+      = timings.toTick<electronics_tick>(timings.TriggerTime());
+    electronics_time_ticks const beforeTicks
+      { -int(detInfo.ReadOutWindowSize()) };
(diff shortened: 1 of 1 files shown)


larsim/EventGenerator/RadioGen_module.cc
```diff
@@ -160,32 +273,93 @@ namespace {
...
+    auto const Bq = pset.get< std::vector<double> >("BqPercc");
+    auto t0 = pset.get< std::vector<double> >("T0", {});
+    auto t1 = pset.get< std::vector<double> >("T1", {});
+    
+    if (fT0.empty() || fT1.empty()) { // better be both empty...
+      assert(fT0.empty() && fT1.empty());
+      auto const times = defaulttimewindow();
+      t0.push_back(times.first);
+      t1.push_back(times.second);
+    }
```
knoepfel: A curious and problematic formulation.  Presumably, you intend something like:
```.cpp
if (size(t0) != size(t1)) {
  throw ...;
}

if (empty(t0)) {
  auto const [start, stop] = defaulttimewindow();
  t0.push_back(start);
  t1.push_back(stop);
}
```

PetrilloAtWork: Fair enough.
The choice was intentional but not well motivated (and probably an abuse of `assert()`).
Pushed a change along the suggested lines.

lgarren:
trigger build

lgarren:
Note that there are problems with the CI dashboard reporting right now.   The actual CI test results are fine.

absolution1:
approve

lgarren:
approve

PR #57, opened 2020-12-09
Fix MCShower's Ancestor Start and End

marcodeltutto:
It looks like there is a typo where the MCShower's ancestor particle Start and End MCSteps are filled. They are filled with the MCShower's mother, instead of the ancestor. This PR fixes this.


lgarren:
trigger build

tomjunk:
Approved.  Well spotted. Looks like the two lines before were cut and pasted but not fully edited.

lgarren:
approve

PR #58, opened 2020-12-23
Add checks for large SCE shifts as a signal to drop charge

tomjunk:
Some charge in ProtoDUNE-SP collects on the electron diverters which poke out into the drift volume.  These changes allow the space charge service in dunetpc to signal to the simulation to drop this charge from the simulation by putting in very large (>1E9 cm) offsets without generating error messages.   It's a bit of a hack -- one might add a flag from the space charge model to say that the charge disappears instead of just being moved, but that would be a much bigger redesign.

### DIFF
--- a/larsim/ElectronDrift/DriftElectronstoPlane_module.cc
+++ b/larsim/ElectronDrift/DriftElectronstoPlane_module.cc
@@ -294,6 +294,11 @@ namespace detsim {
         posOffsetxyz[0] = posOffsets.X();
         posOffsetxyz[1] = posOffsets.Y();
         posOffsetxyz[2] = posOffsets.Z();
+	if (posOffsetxyz[0] < -1E9 || posOffsetxyz[0] > 1E9 ||
+	    posOffsetxyz[1] < -1E9 || posOffsetxyz[1] > 1E9 ||
+	    posOffsetxyz[2] < -1E9 || posOffsetxyz[2] > 1E9) {
+	  continue;
+	}
       }
 
       double avegagetransversePos1 = 0.;
--- a/larsim/ElectronDrift/DriftElectronstoPlane_module.cc
+++ b/larsim/ElectronDrift/DriftElectronstoPlane_module.cc
@@ -59,6 +59,7 @@
 #include "lardataobj/Simulation/SimEnergyDeposit.h"
 #include "larevt/SpaceChargeServices/SpaceChargeService.h"
 #include "larsim/ElectronDrift/ISCalculationSeparate.h"
+#include "larsim/Utils/SCEOffsetBounds.h"
 
 // Framework includes
 #include "art/Framework/Core/EDProducer.h"
@@ -291,14 +292,10 @@ namespace detsim {
       auto const* SCE = lar::providerFrom<spacecharge::SpaceChargeService>();
       if (SCE->EnableSimSpatialSCE() == true) {
         posOffsets = SCE->GetPosOffsets(mp);
+	if (larsim::Utils::SCE::out_of_bounds(posOffsets)) continue;
         posOffsetxyz[0] = posOffsets.X();
         posOffsetxyz[1] = posOffsets.Y();
         posOffsetxyz[2] = posOffsets.Z();
-	if (posOffsetxyz[0] < -1E9 || posOffsetxyz[0] > 1E9 ||
-	    posOffsetxyz[1] < -1E9 || posOffsetxyz[1] > 1E9 ||
-	    posOffsetxyz[2] < -1E9 || posOffsetxyz[2] > 1E9) {
-	  continue;
-	}
       }
 
       double avegagetransversePos1 = 0.;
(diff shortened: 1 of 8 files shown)


larsim/ElectronDrift/DriftElectronstoPlane_module.cc
```diff
@@ -294,6 +294,11 @@ namespace detsim {
         posOffsetxyz[0] = posOffsets.X();
         posOffsetxyz[1] = posOffsets.Y();
         posOffsetxyz[2] = posOffsets.Z();
+	if (posOffsetxyz[0] < -1E9 || posOffsetxyz[0] > 1E9 ||
+	    posOffsetxyz[1] < -1E9 || posOffsetxyz[1] > 1E9 ||
+	    posOffsetxyz[2] < -1E9 || posOffsetxyz[2] > 1E9) {
+	  continue;
+	}
```
knoepfel: Using utility function:

```.cpp
if (out_of_bounds(posOffsetxyz)) {
  continue;
}
```

larsim/ElectronDrift/ShiftEdepSCE_module.cc
```diff
@@ -93,6 +93,14 @@ spacecharge::ShiftEdepSCE::produce(art::Event& e)
...
       posOffsetsStart = sce->GetPosOffsets({edep.StartX(), edep.StartY(), edep.StartZ()});
       posOffsetsEnd = sce->GetPosOffsets({edep.EndX(), edep.EndY(), edep.EndZ()});
+      if (posOffsetsStart.X() < -1E9 || posOffsetsStart.X() > 1E9 ||
+          posOffsetsStart.Y() < -1E9 || posOffsetsStart.Y() > 1E9 ||
+          posOffsetsStart.Z() < -1E9 || posOffsetsStart.Z() > 1E9 ||
+          posOffsetsEnd.X() < -1E9 || posOffsetsEnd.X() > 1E9 ||
+          posOffsetsEnd.Y() < -1E9 || posOffsetsEnd.Y() > 1E9 ||
+          posOffsetsEnd.Z() < -1E9 || posOffsetsEnd.Z() > 1E9) {
+        continue;
+      }
```
knoepfel: Using utility function:

```.cpp
if (out_of_bounds(posOffsetsStart) || out_of_bounds(posOffsetsEnd)) {
  continue;
}
```

knoepfel (commented):
Tom, I suggest writing a utility function that can return a simple `true` or `false` for this type of situation.  For example:

```.cpp
bool out_of_bounds(type_of_offset const& offset)
{
  constexpr double limit{1.e9};
  return std::abs(offset[0]) > limit || 
         std::abs(offset[1]) > limit || 
         std::abs(offset[2]) > limit;
}
```
This will make the code less prone to maintenance errors/typos.

larsim/ElectronDrift/DriftElectronstoPlane_module.cc
tomjunk: Hi Kyle -- Thanks much.  I added a new utility function in larsim/Utils/SCEOffsetBounds.cxx,.h and called it from the four places I had previously hardcoded the bounds check.  I updated the CMakeLists.txt files to link with the larsim_Utils lib where necessary.

lgarren:
trigger build

knoepfel:
Looks good, Tom.  Will approve after L2 approval has been granted.

lgarren:
trigger build

lgarren:
@tomjunk  Would you take a look at the uboone CI failure?  I don't think it is related to this PR, but would like to be sure.

larsim/Utils/SCEOffsetBounds.cxx
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * $Log: $
+ */
+
+// LArSoft
+#include "larsim/Utils/SCEOffsetBounds.h"
+
+bool larsim::Utils::SCE::out_of_bounds(geo::Vector_t const &offset)
+{
+  constexpr double limit{1.e9};
```
absolution1, line 16: This suggestion really doesn't need to hold up the PR (and this suggestion is basically purely stylistic due to there being only one function in the implementation file) but this declaration --could-- be moved outside of the function.

absolution1:
SBND warnings are unrelated, DUNE looks good, uboone has warnings due to sim channel differences.  I'm not a uboone collaborator so cannot really confirm whether this is OK/expected or not...
Can a uboone person confirm that the warnings are benign?

lgarren:
@hgreenlee  would you please comment?

larsim/Utils/SCEOffsetBounds.cxx
tomjunk: Yes, if we need to expose that limit number elsewhere we could define it there.  The intended use is in fact for code in dunetpc to set space charge offsets that exceed the limit so that charge is dropped in the simulation, and making sure that the large value designed to exceed the limit will continue to do so in the future is a reason to expose that declaration somewhere else, but I don't see the need for this number to change.

lgarren:
Tom has checked the uboone CI.  The warnings are unrelated to this PR.

lgarren:
approve

PR #63, opened 2021-02-19
Add DoFastComponent option to PDFastSimPAR

aihimmel:
Add a new fhicl option to turn on and off the Early (singlet) light component as well as the Late (triplet) light component. This is particularly helpful in the context of Xe doping simulation which often has only a late light component. The option has a default value of True for backwards compatibility, making this a non-breaking change.

### DIFF
--- a/larsim/PhotonPropagation/PDFastSimPAR_module.cc
+++ b/larsim/PhotonPropagation/PDFastSimPAR_module.cc
@@ -263,6 +263,7 @@ namespace phot {
       using ODP = fhicl::OptionalDelegatedParameter;
 
       fhicl::Atom<art::InputTag> SimulationLabel  { Name("SimulationLabel"),  Comment("SimEnergyDeposit label.") };
+      fhicl::Atom<bool>          DoFastComponent  { Name("DoFastComponent"),  Comment("Simulate slow scintillation light, default true"), true };
       fhicl::Atom<bool>          DoSlowComponent  { Name("DoSlowComponent"),  Comment("Simulate slow scintillation light") };
       fhicl::Atom<bool>          DoReflectedLight { Name("DoReflectedLight"), Comment("Simulate reflected visible light") };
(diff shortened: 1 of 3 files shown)


lgarren:
trigger build

knoepfel:
This PR should be approved pending L2 approval.

lgarren:
Issue with CI fixed

lgarren:
trigger build

lgarren:
The unrelated problem with the dune CI builds has been fixed.  CI results here:
https://dbweb0.fnal.gov/LarCI/app/ns:DUNE/view_builds/index?offset=0&builds=dune_ci/10977&builds=

larsim/PhotonPropagation/PDFastSimPAR_module.cc
```diff
@@ -263,6 +263,7 @@ namespace phot {
       using ODP = fhicl::OptionalDelegatedParameter;
 
       fhicl::Atom<art::InputTag> SimulationLabel  { Name("SimulationLabel"),  Comment("SimEnergyDeposit label.") };
+      fhicl::Atom<bool>          DoFastComponent  { Name("DoFastComponent"),  Comment("Simulate slow scintillation light, default true"), true };
```
absolution1, line 266: Would experiment code fall over if the default value was removed?
If not, it might be worth considering its removal as its easier to catch FCL errors when default values aren't used.

absolution1:
I had one minor question about default value usage but it's entirely optional so I'll stamp L2 now.

absolution1:
approve

larsim/PhotonPropagation/PDFastSimPAR_module.cc
aihimmel: Yea, I included the default to maintain current behavior and keep the change non-breaking. I think the potential harm is small since I think DUNE are the only ones looking seriously at Xe doping, and that's the only circumstance I know of where it makes sense to turn off the fast component (which I think is why it wasn't an option previously).

lgarren:
approve

PR #67, opened 2021-04-09
Update the larsim interface to MARLEY for compatbility with v1.2.0.

sjgardiner:
NOTE: These changes require upgrading to the marley v1_2_0 ups product. I've made a draft of the necessary changes to the build-framework scripts (`build-framework-marley-ssi-build`) using the same branch name (`feature/gardiner-marley-v120`).

Eliminate the ugly hacks used to convert FHiCL into MARLEY's native JSON format. Instead, use a new class derived from `fhicl::ParameterSetWalker` to manage the conversion. Forego FHiCL validation of the generator settings in favor of letting MARLEY handle that itself. Adjust prodmarley.fcl so that it is experiment-agnostic.

### DIFF
--- a/larsim/EventGenerator/MARLEY/CMakeLists.txt
+++ b/larsim/EventGenerator/MARLEY/CMakeLists.txt
@@ -14,6 +14,7 @@ art_make(LIB_LIBRARIES
            ROOT::Geom
            ROOT::Physics
            ${ART_UTILITIES}
+           ${FHICLCPP}
          MODULE_LIBRARIES
            larsim_EventGenerator_MARLEY
            larcoreobj_SummaryData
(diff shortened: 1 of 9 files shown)


lgarren:
@sjgardiner  Um, not happy with the new build of marley.  Libraries should be in a standard lib directory and source code should be in a source or other directory.  Presumably there is an install option.

lgarren:
@sjgardiner   Oh, for ....   This has an in source build and you expect to use it from inside the source???   That is very much not recommended practice.   I think a proper install can resolve this.

lgarren:
@sjgardiner Looking good now that I've sorted the marley install step.  Will get a CI going as soon as we have marley v1_2_0 on cvmfs.

sjgardiner:
Hi @lgarren, sorry for the admittedly evil build and thanks for sorting things out so quickly. I'll be happy to help out if there are any hiccups in the next stage.

lgarren:
trigger build

absolution1:
Looks good to me!  Always nice to see std::reference_wrapper in use!

absolution1:
approve

lgarren:
approve

PR #68, opened 2021-06-09
Hybrid model for Light simulation UGR

claudiaalvgar:
Pull Request introducing changes needed for using the new Hybrid model for light propagation for the new larg4 in LArSoft. The idea of this model is to use the Semi-Analytic model for propagating photons inside the active volume and a scaled-down Optical-Library for the LAr volume outside the TPC. This way, we are able to have a complete light simulation by having into account the whole cryostat volume.
For using this model some changes are required in larsim. As a summary, our changes are the following: 
- Our proposal is to remove the functions isScintInActiveVolume and extractActiveVolumes from PDFastSimPAR_module. The reason for that is because, in the new larg4, we need them at IonAndScint stage where we want to decide which propagation model to use, via configuration fhicl file, where we specify the volumes where we want to use one model or the other (Semi-Analytic or OpticalLibrary). These functions are needed in ISCalcCorrelated module, in particular, in the function EFieldAtStep() where we need to distinguish between active and non-active volume in order to set a different electric field value inside and outside the active volume. Outside the TPC active volume we need to be able to set to zero (at 1st approach) the EField by using the function isScintInActiveVolume. 
- We propose to create a new class containing these two functions in larsim/IonizationScintillation/, called ISTPC. The reason to create the new class is in order to avoid repeating the functions in both modules (Correlated and PAR), this way both modules are able to use the functions by inheriting the new class.
- Also,  the object “fActiveVolumes" is common to both modules, so it is also included in the new class. 
- Finally in this proposal we have also included the most updated version of PDFastSimPVS_module, provided by Mu Wei which stores reflected light when using the Optical-Library module.  

Your comments will be very welcome: @pgreen135 @aihimmel


aihimmel:
The principle here seems totally sound to me -- let's definitely move common code to a more generic location so multiple modules can use it. However, using multiple inheritance seems like overkill, and might cause some ART problems down the road. (I guess the L1 managers can comment on that point.) It seems like a singleton class (since "Active Volumes" should be the same throughout the job) would do the job and be more robust.

pgreen135:
The changes from the light simulation side look reasonable to me, I will let others comment though re Alex's concern about the inheritance structure.

@claudiaalvgar I triggered the CIs from sbndcode and this currently fails to build in c7 -- looks like an unused variable, see https://dbweb8.fnal.gov:8443/LarCI/app/build_detail/phase_details?build_id=lar_ci/11536&platform=Linux%203.10.0-1160.21.1.el7.x86_64&phase=build&buildtype=slf7%20c7:prof

aihimmel:
Maybe I can call in @knoepfel for a more definitive comment on the multiple inheritance implementation?

knoepfel:
I commend you, @claudiaalvgar, for recognizing that code should not be duplicated when the same functionality is needed in multiple places.  I do agree with @aihimmel that multiple inheritance is not the best route to go here.  But I will comment more next week on some alternatives that still give the desired improvements without introducing unnecessary entanglements through inheritance.  Stay tuned!

larsim/IonizationScintillation/ISCalcCorrelated.cxx
```diff
@@ -130,13 +138,22 @@ namespace larg4 {
...
+     //electric field outside active volume set to zero                                                                                                                                                      
+     if(isScintInActiveVolume(edep.MidPoint())==false) {
+       efield = 0.0;
+       return efield;
+     }
+     else{ //electric field in active volume                                                                                                                                                                 
+       if (!fSCE->EnableSimEfieldSCE()) return efield;
+       auto const eFieldOffsets = fSCE->GetEfieldOffsets(edep.MidPoint());
+       return efield * std::hypot(1 + eFieldOffsets.X(), eFieldOffsets.Y(), eFieldOffsets.Z());
+     }
```
knoepfel: This body of code can be shortened to:

```cpp
if (!isScintInActiveVolume(edep.MidPoint())) return 0.;
if (!fSCE->EnableSimEfieldSCE()) return efield;       

auto const eFieldOffsets = fSCE->GetEfieldOffsets(edep.MidPoint());
return efield * std::hypot(1 + eFieldOffsets.X(), eFieldOffsets.Y(), eFieldOffsets.Z());
```

larsim/IonizationScintillation/ISTPC.cxx
```diff
@@ -0,0 +1,67 @@
...
+#include "larsim/IonizationScintillation/ISTPC.h"
+#include "larcore/CoreUtils/ServiceUtil.h"
+
+namespace larg4 {
+  //----------------------------------------------------------------------------                 
+  ISTPC::ISTPC()
+  {
+    std::cout << "IonizationAndScintillation/ISTPC Initialize." << std::endl;
+    std::cout << "Initializing the geometry of the detector." << std::endl;
+    geo::GeometryCore const& geom = *(lar::providerFrom<geo::Geometry>());
```
knoepfel: Instead of calling `lar::providerFrom<geo::Geometry>()` inside of this class (thus introducing a dependency on the *art* framework), it would be better to pass it into the class through the constructor.  This would then give you a very clean initialization of the class:

```cpp
ISTPC::ISTPC(geo:GeometryCore const& geom) 
  : fActiveVolumes{extractActiveVolumes(geom)}
{ ... }
```

If possible, also avoid directly logging to `std::cout`, which cannot be easily suppressed by users.  It is better to use the message-facility functionality:

```cpp
mf::LogTrace("IonAndScint") << "IonizationAndScintillation/ISTPC Initialize.\n"
                            << "Initializing the geometry of the detector.";
```

larsim/IonizationScintillation/ISTPC.h
```diff
@@ -0,0 +1,34 @@
...
+
+#ifndef IS_ISTPC_H
+#define IS_ISTPC_H
+
+
+#include "larcore/Geometry/Geometry.h"
+#include "larcorealg/CoreUtils/counter.h"
+#include "larcorealg/CoreUtils/enumerate.h"
+
+#include "messagefacility/MessageLogger/MessageLogger.h"
```
knoepfel: These headers should be moved to the .cxx file, which is where they are required.  Otherwise, all downstream code gains a dependency on them, and that's probably not desired.

larsim/IonizationScintillation/ISTPC.h
```diff
@@ -0,0 +1,34 @@
...
+
+    ISTPC();
+    std::vector<geo::BoxBoundedGeo> fActiveVolumes;
+
+     bool isScintInActiveVolume(geo::Point_t const& ScintPoint);
+ 
+    static std::vector<geo::BoxBoundedGeo> extractActiveVolumes(geo::GeometryCore const& geom);
+ 
+      
+  };
```
knoepfel: I would adjust the layout of this class.  The only functionality that is required for downstream code is the `isScintInActiveVolume(...)` function.  To gain maximum flexibility for future development, encapsulate everything else as `private`:

```cpp
class ISTPC {
public:
  explicit ISTPC(geo::GeometryCore const& geom);
  bool isScintInActiveVolume(geo::Point_t const& ScintPoint) const;

private:
  std::vector<geo::BoxBoundedGeo> fActiveVolumes;
  static std::vector<geo::BoxBoundedGeo> extractActiveVolumes(...);
};

File: larsim/IonizationScintillation/ISCalcCorrelated.h [hunk 5]
```diff
@@ -31,7 +32,7 @@
 #include "messagefacility/MessageLogger/MessageLogger.h"
 
 namespace larg4 {
-  class ISCalcCorrelated : public ISCalc {
+  class ISCalcCorrelated : public ISTPC, public ISCalc {
```
knoepfel: See comment below regarding multiple inheritance.

larsim/PhotonPropagation/PDFastSimPAR_module.cc
```diff
@@ -252,7 +251,7 @@ namespace {
 } // namespace
 
 namespace phot {
-  class PDFastSimPAR : public art::EDProducer {
+  class PDFastSimPAR : public larg4::ISTPC, public art::EDProducer {
```
knoepfel: In this case, I would not choose multiple inheritance.  Although inheriting from `ISTPC` can shorten the amount of typing, it also complicates the class hierarchy and potentially introduces entanglements that can be difficult to resolve later on.  Instead, I encourage a private data member of this class that serves the same role:

```cpp
class PDFastSimPAR : public art::EDProducer {
public:
  ...
private:
  larg4::ISTPC fISTPC;
  ...
};
```

The `isScintInActiveVolume(...)` calls will then turn into `fISTPC.isScintInActiveVolume(...)`, which is a little more typing, but it avoids the class-layout complications introduced through inheritance.

I would make this change to the `ISCalcCorrelated` class also.

larsim/PhotonPropagation/PDFastSimPVS_module.cc
```diff
@@ -3,7 +3,7 @@
 // Plugin Type: producer
 // File:        PDFastSimPVS_module.cc
 // Description:
-// - acts on sim::SimEnergyDeposit from LArG4Main,
+// - acts on sim::SimEnergyDeposit from LArG4Main, 
 // - simulate (fast, photon visibility service) the OpDet response to optical photons
 // Input: 'sim::SimEnergyDeposit'
 // Output: 'sim::OpDetBacktrackerRecord'
```
knoepfel, line 9: I will rely on others to make provide feedback on the changes here in this file.  my only comment is there seem to be some whitespace-only changes, which make it difficult to review the meaningful changes.  This usually happens when you don't have the latest changes from the `develop` branch.

knoepfel (commented):
Thanks for the PR, Claudia.  My review covers mostly design issues--keeping data members private where possible, avoiding multiple inheritance, etc.  Happy to follow up on anything.

claudiaalvgar:
List of modificatios (following @knoepfel suggestions):
-Updated constructor for ISTPC class, receiving geom parameter.
-Moved includes from ISTPC.h to ISTPC.cxx where they are used.
-Removed ISTPC inheritance creating an ISTPC object (ISTPC fISTPC;) in ISCalcCorrelated and in PDFastSimPAR_module to access functions like: fISTPC.isScintInActiveVolume and fISTPC.extractActiveLArVolume.
-Initialization of fISTPC object in ISCalcCorrelated and in PDFastSimPAR_module constructor.
-Change the name of extractActiveVolumes to extractActiveLArVolume.
-Updated EFieldAtStep function in ISCalcCorrelated.cxx.
-Removed unused variable (nion) of PDFastSimPVS_module.

knoepfel (commented):
Looks good, Claudia.  This will receive L1 approval after L2 approval has been granted.

knoepfel:
trigger build

lgarren:
trigger build

larsim/IonizationScintillation/ISCalcCorrelated.cxx
```diff
@@ -65,6 +66,9 @@ namespace larg4 {
     double EFieldStep = EFieldAtStep(detProp.Efield(), edep);
     double recomb = 0.;
 
+    //calculate recombination survival fraction value inside, otherwise zero
+
+    if(EFieldStep > 0) {
```
absolution1, line 71: Pure style comment but I'm guessing the code enclosed in this if statement should be indented?

absolution1 (approved):
I had a quick look through and only a small stylistic thing jumped out at me.  
I'll leave @aihimmel to grant L2 approval in this case as he's the expert here.

aihimmel:
As I mentioned above, I think the idea here is a very good one, and it looks like the implementation concern I raised was addressed, so I'm happy for this to proceed. However, I'm no longer an L2 manager so I cannot formally approve.

knoepfel:
trigger build

lgarren:
This PR was approved at the coordination meeting pending the changes which have gone in, so I think we're good once the CI tests finish.  Thanks @aihimmel and @absolution1

lgarren:
approve

PR #76, opened 2021-10-19
Support multiple larg4 instances

marcodeltutto:
As shown in issue [26246](https://cdcvs.fnal.gov/redmine/issues/26249), running multiple larg4 instances requires some changes to larg4 and larsim. The larg4 changes are implements in PR LArSoft/larg4#30, while this PR deals with the changes in larsim, specifically:
- Adds fhich input parameter `InputModuleLabels` to the `IonAndScint` module, so one can run `IonAndScint` multiple times with different input collections.
- Adds more fhicl configurations to `MergeSimSources`, so one can select what input collections to merge.
- The default behavior of both modules should be unchanged.

### DIFF
--- a/larsim/IonizationScintillation/IonAndScint_module.cc
+++ b/larsim/IonizationScintillation/IonAndScint_module.cc
@@ -100,8 +100,7 @@ namespace larg4 {
       fSelector = s;
     }
 
-    // mf::LogDebug("IonAndScint") << "Selecting input collections based on \n"
-    std::cout << "Selecting input collections based on \n"
+    mf::LogDebug("IonAndScint") << "Selecting input collections based on \n"
                                 << fSelector.print(" ") << std::endl;
 
 
--- a/larsim/IonizationScintillation/IonAndScint_module.cc
+++ b/larsim/IonizationScintillation/IonAndScint_module.cc
@@ -43,6 +43,7 @@
 #include <string>
 
 using std::string;
+using SimEnergyDepositCollection = std::vector<sim::SimEnergyDeposit>;
 
 namespace larg4 {
   class IonAndScint : public art::EDProducer {
@@ -53,7 +54,15 @@ namespace larg4 {
     void endJob() override;
 
   private:
-    art::InputTag calcTag; // name of calculator: Separate, Correlated, or NEST
+    std::vector<art::Handle<SimEnergyDepositCollection>> inputCollections(art::Event const& e) const;
+
+    // name of calculator: Separate, Correlated, or NEST
+    art::InputTag calcTag;
+
+    // Input tags used to specify which SimEnergyDeposit collections to use
+    // If empty, this module uses all the available collections in the event
+    std::vector<art::InputTag> inputCollectionTags;
+
     std::unique_ptr<ISCalc> fISAlg;
     CLHEP::HepRandomEngine& fEngine;
     string Instances;
@@ -65,6 +74,7 @@ namespace larg4 {
   IonAndScint::IonAndScint(fhicl::ParameterSet const& pset)
     : art::EDProducer{pset}
     , calcTag{pset.get<art::InputTag>("ISCalcAlg")}
+    , inputCollectionTags{pset.get<std::vector<art::InputTag>>("InputCollections", {})}
     , fEngine(art::ServiceHandle<rndm::NuRandomService>()
                 ->createEngine(*this, "HepJamesRandom", "NEST", pset, "SeedNEST"))
     , Instances{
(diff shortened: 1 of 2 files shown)


lgarren:
trigger build

knoepfel:
@marcodeltutto, the PR is very well done.  I have one question, though.  The behavior of the selector is a little different from what I implemented for https://github.com/LArSoft/larg4/pull/30: 

- For `larg4`, if you specify `input_mct_collections`, each collection must be present.
- In this PR, not all collections must be present (due to the Boolean OR of the selectors)

It this okay?  Or do you think the behaviors from both packages should be unified?

marcodeltutto:
Hi @knoepfel, yes that's a good point, thanks! It would probably be better to require all collections to be present. I can't find an easy way to do this though. Switching from boolean OR to AND won't work as it will require a module label to be equal to two strings at the same time. If we use `getHandle<T>(InputTag const& tag)` as you did in your PR, do I understand correctly that this will either grab a particular instance with `label:instance` or the latest instance placed in the event with just `label`? 

If we want to keep the current design of this module, we want all instances given a specified set of module labels. I can always call `getMany(selector)` multiple times with each of the specified labels and then merge the results, but I was hoping for something cleaner.

knoepfel:
@marcodeltutto, I used the `InputTag` interface because it is more flexible in its specification--it takes up to three fields, separated by a colon delimiter:

```c++
art::InputTag const tag{"moduleLabel:optionalInstanceName:optionalProcessName"};
```
- The `moduleLabel` is always required.
- The `optionalInstanceName` is only required to retrieve a product with a non-empty instance name, or when retrieving a product from a particular process.
- The `optionalProcessName` is only required to retrieve a product created in a specific process.

For example, in each of the following invocations, the product was created in a module with the label "myLabel"

```c++
// Product retrieval call                  C++ type | Instance name | Process
e.getHandle<int>("myLabel");               // int   | ""            | most recent
e.getHandle<int>("myLabel:");              // int   | ""            | most recent
e.getHandle<int>("myLabel::");             // int   | ""            | most recent
e.getHandle<int>("myLabel:myInst");        // int   | myInst        | most recent
e.getHandle<int>("myLabel:myInst:");       // int   | myInst        | most recent
e.getHandle<int>("myLabel::myProc");       // int   | ""            | myProc
e.getHandle<int>("myLabel:myInst:myProc"); // int   | myInst        | myProc
```

So, I *think* there is enough flexibility in the configuration as long as the strings specified in the configuration convert cleanly to the `art::InputTag` constructor as described above.

**Bottom-line**: Kudos on using *art*'s selector technology, but I don't think you need to in this case.  Just a simple loop over `art::Event::getHandle<T>(InputTag const&)` calls should be sufficient.

knoepfel:
@marcodeltutto, sorry, I submitted the comment too soon.  I've updated the previous comment: https://github.com/LArSoft/larsim/pull/76#issuecomment-947981679.

marcodeltutto:
Thanks @knoepfel! I see what you mean, but in this case we have several instances per module, and we want 
to be able to select them all by only provding the module label. For example, we may have

```ruby
module1:instance1
module1:instance2

module2:instance1
module2:instance2
module2:instance3
```

And we want to grab them all by specifing `InputCollections: ["module1", "module2"]`. For this we would need a getMany()
and then I believe a selector to specify the module. But I can't find a simple selector that can select all instances of
a specified list of modules. That's why I went for the OR of `art::ModuleLabelSelector`, as this would do the job
but has the downside that if one module is missing we won't get an error.

The alternative to me would be to do something like

```cpp
    std::vector<art::Handle<SimEnergyDepositCollection>> result;
    for (auto const & module : fInputModuleLabels) {
      auto handels = e.getMany<SimEnergyDepositCollection>(art::ModuleLabelSelector(module));
      if (empty(handels)) {
        throw art::Exception(art::errors::ProductNotFound)
          << "IonAndScint module cannot find any SimEnergyDeposits with module label "
          << module << " as requested in InputModuleLabels. \n";
      }
      result.insert(result.end(), handels.begin(), handels.end());
    }
```

Please let me know if I am missing something obvious here!

marcodeltutto:
@knoepfel I update this PR with commit c2adcc5, this ensures that products with specified labels have to be found in the event.

absolution1:
approve

knoepfel:
Well done, @marcodeltutto.  I think you chose the best solution available.

knoepfel:
Approve

lgarren:
trigger build

knoepfel:
@marcodeltutto, I understand what is going on.  There are parts of the `larg4Main` module that can only be called/constructed/initialized/destroyed once, and other parts that can be used multiple times.  Now that `artg4tk` allows you to specify which collections to use, I need to tease apart those details.  I should have something ready by sometime next week.  Thanks for your patience.

marcodeltutto:
Thank you @knoepfel, much appreciated!

PR #81, opened 2021-11-17
Add Option to Merge AuxDetHits in MergeSimSources Module

marcodeltutto:
This PR adds the option to merge `sim::AuxDetHits` from different streams in the `MergeSimSources` modules. This is needed when we run more than one instance of LArG4 and we then want to merge the output.

The default behavior of `MergeSimSources` is unchanged, this PR only adds an additional option that right now we need in SBND.

### DIFF
--- a/larsim/MergeSimSources/MergeSimSources_module.cc
+++ b/larsim/MergeSimSources/MergeSimSources_module.cc
@@ -34,6 +34,7 @@
 #include "lardataobj/Simulation/SimPhotons.h"
 #include "lardataobj/Simulation/SimChannel.h"
 #include "lardataobj/Simulation/SimEnergyDeposit.h"
+#include "lardataobj/Simulation/AuxDetHit.h"
 #include "nusimdata/SimulationBase/MCTruth.h"
 #include "nusimdata/SimulationBase/MCParticle.h"
 #include "larcorealg/CoreUtils/enumerate.h"
@@ -103,6 +104,19 @@ public:
       std::vector<std::string>{ "TPCActive", "Other" } // default
     };
 
+    fhicl::Atom<bool> FillAuxDetHits {
+      fhicl::Name{ "FillAuxDetHits" },
+      fhicl::Comment
+        { "whether to merge aux det hit collections" },
+      false // default
+    };
+
+    fhicl::Sequence<std::string> AuxDetHitsInstanceLabels {
+      fhicl::Name{ "AuxDetHitsInstanceLabels" },
+      fhicl::Comment{ "labels of AuxDetHits collections to merge" },
+      std::vector<std::string>{ "LArG4DetectorServicevolAuxDetSensitiveCRTStripY", "Other" } // default
+    };
+
   }; // struct Config
 
   using Parameters = art::EDProducer::Table<Config>;
@@ -124,6 +138,8 @@ private:
   bool                       const fFillAuxDetSimChannels;
   bool                       const fFillSimEnergyDeposits;
   std::vector<std::string>   const fEnergyDepositionInstances;
+  bool                       const fFillAuxDetHits;
+  std::vector<std::string>   const fAuxDetHitsInstanceLabels;
 
   static std::string const ReflectedLabel;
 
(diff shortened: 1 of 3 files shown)


lgarren:
trigger build

larsim/MergeSimSources/MergeSimSources_module.cc
```diff
@@ -260,6 +290,11 @@ void sim::MergeSimSources::produce(art::Event & e)
     for (auto const i [[maybe_unused]]: util::counter(fEnergyDepositionInstances.size()))
       edepCols.push_back(std::make_unique<std::vector<sim::SimEnergyDeposit>>());
   } // if
+  std::vector<std::unique_ptr<std::vector<sim::AuxDetHit>>> auxdethitCols;
+  if (fFillAuxDetHits) {
+    for (auto const i [[maybe_unused]]: util::counter(fAuxDetHitsInstanceLabels.size()))
+      auxdethitCols.push_back(std::make_unique<std::vector<sim::AuxDetHit>>());
+  } // if
```
knoepfel: This is a bit awkward.  I suggest

```c++
using aux_det_hits_t = std::vector<sim::AuxDetHit>;

std::vector<aux_det_hits_t> auxdethitCols;
if (fFillAuxDetHits)
  auxdethitCols.resize(fAuxDetHitsInstanceLabels.size());
}    
```
and then...

larsim/MergeSimSources/MergeSimSources_module.cc
```diff
@@ -331,6 +366,18 @@ void sim::MergeSimSources::produce(art::Event & e)
...
 
+
+    if (fFillAuxDetHits) {
+      for (auto const& [ auxdethit_inst, auxdethitCol ]
+        : util::zip(fAuxDetHitsInstanceLabels, auxdethitCols))
+      {
+        art::InputTag const auxdethit_tag { input_label.label(), auxdethit_inst };
+        auto const& input_auxdethit
+          = e.getProduct<std::vector<sim::AuxDetHit>>(auxdethit_tag);
+        MergeUtility.MergeAuxDetHits(*auxdethitCol, input_auxdethit, i_source);
```
knoepfel: ... I suggest
```c++
MergeUtility.MergeAuxDetHits(auxdethitCol, 
                             e.getProduct<aux_det_hits_t>(auxdethit_tag), 
                             i_source);
```
and then...

larsim/MergeSimSources/MergeSimSources_module.cc
```diff
@@ -360,6 +407,14 @@ void sim::MergeSimSources::produce(art::Event & e)
     } // for
   } // if fill energy deposits
 
+  if(fFillAuxDetHits) {
+    for (auto&& [ auxdethit_inst, auxdethitCol ]
+      : util::zip(fAuxDetHitsInstanceLabels, auxdethitCols))
+    {
+      e.put(std::move(auxdethitCol), auxdethit_inst);
```
knoepfel: ... I suggest
```c++
e.put(std::make_unique<aux_det_hits_t>(move(auxdethitCol)), auxdethit_inst);
```
Should provide similar efficiency.

knoepfel (commented):
PR looks pretty good, @marcodeltutto.  I suggest using `std::vector<std::vector<T>>` instead of `std::vector<std::unique_ptr<std::vector<T>>>`.   It is a little easier to reason about and the translation from `std::vector<T>` to `std::unique_ptr<std::vector<T>>` can happen at the `e.put` stage with reasonable efficiency.

larsim/MergeSimSources/MergeSimSources_module.cc
marcodeltutto: Done with commit db484f9

larsim/MergeSimSources/MergeSimSources_module.cc
marcodeltutto: Done with commit db484f9

larsim/MergeSimSources/MergeSimSources_module.cc
marcodeltutto: Done with commit db484f9

marcodeltutto:
Thanks @knoepfel! I wrote it that way to keep it symmetric with the other "MergeSimEnergyDeposit" code that was already there. But I agree the way you suggest is better, so I made to changes to both "MergeAuxDetHits" and "MergeSimEnergyDeposit". Thanks!

knoepfel:
trigger build

lgarren:
approve

PR #82, opened 2021-11-22
Map SimChannels by Cryostat only to prevent extra SimChannels from be…

JamesJieranShen:
Map SimChannels by Cryostat only to prevent extra SimChannels from being created when a channel is associated with multiple TPCs

Part of redmine issue https://cdcvs.fnal.gov/redmine/issues/26453

### DIFF
--- a/larsim/ElectronDrift/SimDriftElectrons_module.cc
+++ b/larsim/ElectronDrift/SimDriftElectrons_module.cc
@@ -134,7 +134,7 @@ namespace detsim {
     typedef std::map<raw::ChannelID_t, ChannelBookKeeping> ChannelMap_t;
 
     // Array of maps of channel data indexed by [cryostat,tpc]
-    std::vector<std::vector<ChannelMap_t>> fChannelMaps;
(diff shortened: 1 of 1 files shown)


absolution1:
trigger build

knoepfel:
Thanks @JamesJieranShen--the coding side looks fine.  I'll let Dom comment on potential physics implications, if there are any.

absolution1:
I think the physics implications are good.  This only affects experiments that are
 - using the refactored LArG4 with the provided 'elecDrift' drift simulation
 - using channels with shared wires
 
 So, the only affected customers are DUNEFD (but not in the main simulation chain) and ICARUS (who are moving to wirecell).
 
 I have seen event displays showing the fix working correctly so I think we're all set.

absolution1:
approve

lgarren:
trigger build
(picking up uboone)

lgarren:
uboone unit tests OK in local build

lgarren:
approve

PR #84, opened 2021-12-06
Semi-analytical light simulation refactoring

### LINKED ISSUE #163: OpT0Finder Is SegFaulting in v09_28_01
in SBNDCode v09_28_01 it looks like OpT0Finder has a few issues which it either seg faults or gives the following error:
`Error: Invalid optical detector type. 0 = rectangular, 1 = dome, 2 = disk. Or corrections for chosen optical detector type missing.` before crashing

pgreen135:
This PR restructures PDFastSimPAR to factor out the semi-analytical model number of photons and propagation times calculations into two new classes, SemiAnalyticalModel.h/cpp and PropagationTimeModel.h/cpp. This allows these models to be accessed from other modules/algorithms by creating separate instances of these classes, rather than only being able to access them through the PDFastSimPAR producer. For the propagation time model, this allows it to be used in conjunction with optical libraries in PDFastSimPVS as was the case in the legacy LArG4 (required for the hybrid model in SBND, resolving https://github.com/SBNSoftware/sbndcode/issues/219, and will be required by ICARUS once they transition to the refactored LArG4). For the number of photons model, this allows it to be used during flash matching (needed to resolve https://github.com/SBNSoftware/sbndcode/issues/163). This hopefully should also make the code more easily maintainable by making PDFastSimPAR less monolithic. 

Changes:
- Moved semi-analytical model (number of photons) functions to new class, SemiAnalyticalModel.h/cpp, and removed corresponding code from PDFastSimPAR. Altered to return visibilities analogous to the photon library rather than the number of photons to integrate more easily with the fast/slow photon split introduced for Xenon-doped simulations.
- Moved propagation time functions to new class, PropagationTimeModel.h/cpp, and removed corresponding code from PDFastSimPAR.
- Implemented propagation time model calculations into PDFastSimPVS (using the PropagationTimeModel class), to allow use when predicting number of photons using a library restoring the legacy LArG4 functionality.
- Restructured PDFastSimPVS to reduce code repetition and fix a bug that prevented reflected/visible photons from being saved if no direct/VUV photons are detected. The structure now matches that in PDFastSimPAR. 
- Minor optimisations and cleanup, resolving excessive vector construction/destruction that was unintentionally introduced in the recent updates to the semi-analytical model for DUNE vertical drift. This speeds up the reflected/visible light number of photons calculations significantly. I also ran some simple profiling and did not spot any other obvious issues.

The changes are not expected to alter the light seen in any of the detectors using these models, other than the minor changes mentioned above the functions are unchanged from the PDFastSimPAR implementations. I have tested this in SBND, DUNE-HD and DUNE-VD and they appear to be working as expected.


lgarren:
@pgreen135  There are many whitespace  only changes included in this PR.  Unfortunately, that makes it difficult to review.   Is there a chance that you can separate the whitespace only changes out?

pgreen135:
@lgarren it looks like some white space changes were picked up specifically in the PDFastSimPVS module -- is there an easy way to undo these? 
For PDFastSimPAR, the changes are as a result of these variables/functions being moved to the new classes, rather than white-space differences as far as I can see in the diff.

pgreen135:
I think many of these are related to mixing tabs and spaces -- I'll try to undo these (replace everything with spaces)

knoepfel:
@pgreen135, there are not many good ways of doing this.  We should find a way to adopt `clang-format` usage so this is never a problem--I'll bring this up with @erica-snider again.  I appreciate your efforts to undo the space changes.  If you run into problems doing that, let me know and I'll see what I can do on my end.

pgreen135:
I've removed all tabs, so at least the crazy indentation is gone. However, the white space differences remain (and may actually be worse now...). Many of these differences are from tidying up incorrect indents as I was doing the restructuring (so very hard to undo), but I agree that makes this even more difficult to review.

knoepfel:
Thanks @pgreen135.  I'll take a look.

lgarren:
trigger build

absolution1:
approve

lpaulucc (commented):
Although the specific changes in the code are hard to spot, it seems to be no concerns to the modifications that are proposed, and that they should be approved in order to solve SBND related issues. Tests have been performed in all codes that currently use the semi-analytical light sim. Just a few questions for @pgreen135: 
1. The code specifically asked about the cathode position. In the current FD2-VD geometry the cathode is not explicitly there yet. Since tests were made, I assume this is not an issue but just checking; 
2. Transportation times in the hybrid mode are taken from the semi-analytical parameterization even when the optical library is used. I assume this will only be the case for the hybrid mode and that when using the optical library or GAN method alone the code remains as it was before, so no breaking changes here, correct?

pgreen135:
@lpaulucc thanks for taking a look -- 
1. In FD2-VD the configuration runs in "AnodeMode" where it uses the coordinates and dimensions of the anode rather than the cathode for the reflected light calculations. The implementation of this in SemiAnalyticalHits is identical to the previous one in PDFastSimPAR, except for a change to how the different parameter sets are swapped between to avoid excessive vector construction/destruction that I mistakenly introduced in the previous updates for the vertical drift. I've tested this running the standard VD workflow and see no change in the amount of light.
2. The optical libraries do not provide transport times, only visibilities, and currently at least these can only be modelled with the parameterised approach. In the legacy larg4 there was the functionality to use the transport time model along with the optical library (ICARUS currently still does this in legacy, and this is needed for the Hybrid model in SBND), but this was not implemented in the refactored LArG4. This PR restores that functionality when requested in the configuration, experiments running with the optical library only should be unaffected -- so no breaking change. The GAN model is unrelated to PDFastSimPVS/PAR. This change though would also allow the transport time model to be used in conjunction with the GAN model if that is desired.

lpaulucc:
Thanks for the answers, @pgreen135. This all looks fine and I see no reason not to approve this merge.

lgarren:
approve

PR #85, opened 2021-12-21
Feature/andrzejs text file gen

therealendrju:
Add Offset to TextFileGen module This allows controlling it via a script in grind jobs, making running grid jobs easier. 
New parameter defaults to zero, should be invisible to users who don't care about it. 
Restructured the code - moved the event reading into a member function.

### DIFF
--- a/larsim/EventGenerator/TextFileGen_module.cc
+++ b/larsim/EventGenerator/TextFileGen_module.cc
@@ -86,8 +86,8 @@ public:
 
 private:
 
-
-  void ReadNextHepEvt(std::vector<simb::MCParticle> & NextEvent);
+  std::pair<unsigned, unsigned> readEventInfo(std::istream& is);
+  simb::MCTruth  readNextHepEvt();
   unsigned long int fOffset;
   std::ifstream* fInputFile;
   std::string    fInputFileName; ///< Name of text file containing events to simulate
@@ -128,6 +128,15 @@ void evgen::TextFileGen::beginJob()
     throw cet::exception("TextFileGen") << "input text file "
 					<< fInputFileName
 					<< " cannot be read.\n";
+        
+
+  for (unsigned i = 0; i != fOffset; ++i) {
+    auto const [eventNo, nparticles] = readEventInfo(*fInputFile);
+    for (unsigned p = 0; p != nparticles; ++p) {
+       constexpr auto all_chars_until = std::numeric_limits<unsigned>::max();
+       fInputFile->ignore(all_chars_until, '\n');
+    }
+  }                    
                     
 
 }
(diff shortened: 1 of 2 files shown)


knoepfel:
Thanks @therealendrju for the PR.  It's clear what your intention is--however, I think a cleaner implementation can be achieved without having to make as many changes as you did.  The important aspect is to skip the first *n* events--and that's something that can be done in `beginJob` with the following code:

```c++
void beginJob() 
{
  fInputFile = new std::ifstream(fInputFileName.c_str());
  if ( *fInputFile )
     throw cet::exception("TextFileGen") << "input text file "
 					<< fInputFileName
 					<< " cannot be read.\n";

    std::istringstream buffer;
    for (unsigned i = 0; i != fOffset; ++i) {
        std::string line;
        getline(iss, line);
        buffer.str(line);

        // Parse read line for the event number and particles per event
        unsigned event, nparticles;
        buffer >> event >> nparticles;
        for (unsigned p = 0; p != nparticles; ++p) {
            constexpr auto all_chars_until = std::numeric_limits<unsigned>::max();
            iss.ignore(all_chars_until, '\n');
        }
    }
}
```

Then the `produce` function can be how it was before.

therealendrju:
Hi @knoepfel ,
thanks! I agree that moving the initial jumping to the offset to beginJob is definitely a cleaner solution.
But if I use your method, then we effectively have two pieces of code to do the same thing, i.e. read in the number of particles and skip the n-lines. I'd feel better if we had just one piece of code to do it. Can I propose to keep the external function and move its initial reading into the beginJob? Or is there a reason you don't want that code moved out to a function?
thanks!

knoepfel:
@therealendrju, breaking apart the code into more manageable bits (e.g. external functions) is a good thing--I am totally in favor of that!  But I think the granularity of those bits probably needs to be adjusted.  The main concern I have with your solution is that to skip to an offset of *n*, you have to create a vector *n* times, and do a lot of computation that you'll just throw away.  How about this...

1. Create utility to read event number and particles

```c++
std::pair<unsigned, unsigned> readEventInfo(std::istream& is)
{
  std::string line;
  getline(iss, line);
  std::istringstream buffer{line};

  // Parse read line for the event number and particles per event
  unsigned event, nparticles;
  buffer >> event >> nparticles;
  return {event, nparticles};
}
```

2. The `beginJob` function can then look like:
```c++
void beginJob() 
{
  fInputFile = new std::ifstream(fInputFileName.c_str());
  if ( *fInputFile )
     throw cet::exception("TextFileGen") << "input text file "
 					<< fInputFileName
 					<< " cannot be read.\n";

  for (unsigned i = 0; i != fOffset; ++i) {
    auto const [eventNo, nparticles] = readEventInfo(*fInputFile);
    for (unsigned p = 0; p != nparticles; ++p) {
       constexpr auto all_chars_until = std::numeric_limits<unsigned>::max();
       iss.ignore(all_chars_until, '\n');
    }
  }
}
```

3. You can still provide a `readNextHepEvt` function, but I would have it return a `simb::MCTruth` object:
```c++
simb::MCTruth readNextHepEvt() 
{
  // ...
  simb::MCTruth nextEvent;
  auto const [eventNo, nParticles] = readEventInfo(*fInputFile);
  for (unsigned short i = 0; i < nParticles; ++i) {
    ...
    simb::MCParticle part(i, pdg, "primary", firstMother, mass, status);
    ...
    nextEvent.Add(part);
  }
  return nextEvent;
}
```
The `produce` function is then pretty simple:
```c++
void evgen::TextFileGen::produce(art::Event & e)
{
  // check that the file is still good
  if( !fInputFile->good() )
    throw cet::exception("TextFileGen") << "input text file "
					<< fInputFileName
					<< " cannot be read in produce().\n";
   auto truthcol = std::make_unique<std::vector<simb::MCTruth>>();
   truthcol->push_back(readNextHepEvt());
   e.put(move(truthcol));
 }
```
What do you think?

therealendrju:
Hi @knoepfel ,
thanks again. That looks very elegant - I like it. I will try to implement that and get back to you.

cheers,
Andrzej

therealendrju:
Hi @knoepfel, I have implemented your suggestions, 
and cleaned up after. Code builds and works as previously as far as I can tell.
thanks!
Andrzej

knoepfel:
Looks good, @therealendrju.  I'll approve once there's L2 approval.

lgarren:
trigger build

larsim/EventGenerator/TextFileGen_module.cc
```diff
@@ -94,17 +97,19 @@ class evgen::TextFileGen : public art::EDProducer {
 //------------------------------------------------------------------------------
 evgen::TextFileGen::TextFileGen(fhicl::ParameterSet const & p)
   : EDProducer{p}
+  , fOffset{p.get<unsigned long int>("Offset",0)}
```
absolution1: do you need the default value here?  Default values do make it harder to catch fcl mistakes...

therealendrju: that makes it invisible to people who don't want this feature, i.e. people don't need to update their .fcl files if they don't want to.

absolution1: Is there a base FCL to update as that could absorb the change?  Otherwise I might argue that any new parameter is something an end user may want to know about.

knoepfel: @therealendrju, are there any circumstances in which setting the offset value to something other than 0 is *necessary* to get the correct result?  If so, removing the C++ default of 0 (@absolution1's suggestion) may be the best option.

lgarren: > @therealendrju, are there any circumstances in which setting the offset value to something other than 0 is _necessary_ to get the correct result? If so, removing the C++ default of 0 (@absolution1's suggestion) may be the best option.

@therealendrju - comments please?

therealendrju: Hi, sorry. Yes, the value is what steers the behaviour of the module, so if you want to use the new functionality you need to set it to something different than zero. I'll remove the default value and uncomment the .fcl file to provide the zero there.

therealendrju: @lgarren , @absolution1 , @knoepfel   - commit pushed.

lgarren:
trigger build

lgarren:
@therealendrju  the uboonecode CI tests are failing.  The problem may be in ubsim/ubsim/EventGenerator.

lgarren:
@therealendrju nevermind.  The CI test failures exist when this PR is not included.  Sorry for the noise.

absolution1:
approve

lgarren:
approve

PR #86, opened 2022-01-25
Correct IDE energy units

henrylay97:
As requested by @hanswenzel in LArSoft/larg4#34 I am moving the conversion of IDE energy from MeV to GeV further downstream. @ilepetic has confirmed SBND are currently the only users of GenericCRT, so this should only effect our use case.

### DIFF
--- a/larsim/GenericCRT/GenericCRT.cxx
+++ b/larsim/GenericCRT/GenericCRT.cxx
@@ -19,7 +19,7 @@ sim::AuxDetIDE sim::GenericCRTUtility::toAuxDetIDE(const sim::AuxDetHit &InputHi
     sim::AuxDetIDE outputIDE;
 
    outputIDE.trackID		    = InputHit.GetTrackID();
-   outputIDE.energyDeposited	= InputHit.GetEnergyDeposited();
+   outputIDE.energyDeposited	= InputHit.GetEnergyDeposited() / CLHEP::GeV;
    outputIDE.entryX		= InputHit.GetEntryX();
    outputIDE.entryY		= InputHit.GetEntryY();
    outputIDE.entryZ		= InputHit.GetEntryZ();
--- a/larsim/GenericCRT/GenericCRT.cxx
+++ b/larsim/GenericCRT/GenericCRT.cxx
@@ -5,7 +5,7 @@
  * Description:
  * Class with Algorithms to convert sim::AuxDetHits to sim::AuxDetSimChannels
  *
-*/
+ */
 
 
 #include "GenericCRT.h"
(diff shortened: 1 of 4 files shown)


lgarren (commented):
@henrylay97 What are the units here?  If they are already MeV, then this might not give the expected conversion?

henrylay97:
The units of the ``InputHit.GetEnergyDeposited()`` function returns in MeV. The ``CLHEP::GeV`` const has the value of 1000 to allow for conversion from MeV to GeV. I have verified that this works by printing the values before and after as well as the value of ``CLHEP::GeV``. I have also verified that it corrects the issue we were seeing in our detector simulation, further confirming my understanding.

SiPM readout values for SBND's bottom plane (true for all) before the change...
ADCBot
and after the change...
ADCBot

lgarren:
Much thanks @henrylay97   I should have remembered that myself. ;)

lgarren:
trigger build

henrylay97:
@lgarren no problem! Its not exactly intuitive!

knoepfel:
@henrylay97, although your solution may meet the needs of SBND, we would like to retain the generic character of the code by allowing more flexibility in the choice of units.  Is it possible to adjust this PR so that (e.g.) `CLHEP::GeV` could be specified at the function call site, or perhaps via a configuration option?  The default energy unit should remain MeV.

henrylay97:
Hi all, I have updated the PR as requested. I haven't worked with this split module & alg configuration before, so my solution may well need some refining. 

@knoepfel @hanswenzel could I ask why the active choice of changing to MeV has been made here? It seems a little at odds with the use of GeV elsewhere in larsoft code.

hanswenzel:
approve

knoepfel:
@henrylay97, it's rare that all processing steps of a workflow would use consistent units for everything--for some physics processes, MeV is appropriate (like in Geant4); in others, GeV is better (like truth-matching to generated particles).  This is one of those cases.

And thanks for accommodating our request re. the configurability of the scale.  I would like to explore ways to adjust this PR so that

```diff
- EnergyUnitsScale: 1e-3
+ EnergyUnitsScale: "GeV"
```

But I'll have to get back to you in a couple days on that.  We're headed in the right direction!

henrylay97:
@knoepfel hope this is what you were after. [This](https://github.com/LArSoft/larsim/pull/86/commits/03a45244cd6ad019c9c43213d8627e9820bba353) commit should make the change you requested. [This](https://github.com/LArSoft/larsim/pull/86/commits/26dd25eac3b83243a5a25eed9ca9372e5d7ed5b7) commit just cleans up the indentation as it was annoying me a little. Thanks for the clarification on units too!

knoepfel:
trigger build

knoepfel:
@henrylay97, yes, that's a much better solution!  I didn't realize CLHEP supported something like it.  Will approve once the CI tests come back successful.

henrylay97:
@knoepfel no neither did I but a bit of googling always works well haha

lgarren:
There is now a successful uboone CI build: https://dbweb8.fnal.gov:8443/LarCI/app/ns:uBooNE/view_builds/index?offset=0&builds=uboone_ci/7121&builds=

lgarren:
And the dunesw test: https://dbweb8.fnal.gov:8443/LarCI/app/ns:DUNE/view_builds/index?offset=0&builds=dune_ci/12798&builds=

lgarren:
approve

PR #92, opened 2022-05-02
Migrate to Cetmodules 3.12.00 and modern CMake

greenc-FNAL:
- Migrate to using Cetmodules 3.10.00
- Cetmodules -> 3.11.03
- Resolve MIGRATE-ACTION, improve dependencies, plugins and modern CMake use
- Presets generated by Cetmodules 3.12.00
- Add headers target
- Add new target larsim::ScintTimeTool and plugin builder lar::ScintTimeTool


lgarren:
approve

PR #93, opened 2022-05-12
Fix to allow PhotonBacktrackerService to use multiple scintillation labels

ikatza:
This became necessary since the G4 refactoring, where multiple instances of light scintillation are started to handle the various aspects of light simulation.

Care has been taken to ensure backwards compatibility, and I believe that it shouldn't break anything.

I cleaned the code a bit, there are some remaining comments that I haven't removed but I'm happy to chuck them if someone more knowledgable is sure they're not needed anymore.

### DIFF
--- a/larsim/MCCheater/PhotonBackTracker.tcc
+++ b/larsim/MCCheater/PhotonBackTracker.tcc
@@ -17,45 +17,52 @@ namespace cheat{
     void PhotonBackTracker::PrepOpDetBTRs(Evt const& evt)
     {
       if(this->BTRsReady()){ return;}
-      auto const& btrHandle = evt.template getValidHandle < std::vector < sim::OpDetBacktrackerRecord > > (fG4ModuleLabel);
-      //      if(btrHandle.failedToGet()){
-      /*  mf::LogWarning("PhotonBackTracker") << "failed to get handle to     simb::MCParticle from "
-       *              << fG4ModuleLabel
-       *                          << ", return";*/ //This is now silent as it is expected to    happen every generation run. It is also temporary while we wait for
-      /*if( 0 ){ return;} //Insert check for DivRecs here, or don't use validHandle below.
-        auto const& divrecHandle = evt.template getValidHandle <std::vector<sim::OpDetDivRec>>(fWavLabel);
-        if(divrecHandle.failedToGet()){
-        return;
-        }*/
+      const std::vector<art::InputTag> G4ModuleLabels = (fG4ModuleLabels.empty()) ?
+        std::vector{fG4ModuleLabel} : fG4ModuleLabels;
 
-      art::fill_ptr_vector(priv_OpDetBTRs, btrHandle);
-      //art::fill_ptr_vector(priv_DivRecs, divrecHandle);
+      auto compareBTRlambda = [](art::Ptr<sim::OpDetBacktrackerRecord> a,
+                                 art::Ptr<sim::OpDetBacktrackerRecord> b)
+                                {return(a->OpDetNum()<b->OpDetNum());};
 
-      auto compareBTRlambda = [](art::Ptr<sim::OpDetBacktrackerRecord> a, art::Ptr<sim::OpDetBacktrackerRecord> b) {return(a->OpDetNum()<b->OpDetNum());};
-      if (!std::is_sorted(priv_OpDetBTRs.begin(),priv_OpDetBTRs.end(),compareBTRlambda)) 
-        std::sort(priv_OpDetBTRs.begin(),priv_OpDetBTRs.end(),compareBTRlambda);
-      //auto compareDivReclambda = [](art::Ptr<sim::OpDetDivRec> a, art::Ptr<sim::OpDetDivRec> b) {return(a->OpDetNum() < b->OpDetNum());};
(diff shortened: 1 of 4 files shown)


larsim/MCCheater/PhotonBackTracker.tcc
```diff
@@ -17,45 +17,52 @@ namespace cheat{
...
-      /*  mf::LogWarning("PhotonBackTracker") << "failed to get handle to     simb::MCParticle from "
-       *              << fG4ModuleLabel
-       *                          << ", return";*/ //This is now silent as it is expected to    happen every generation run. It is also temporary while we wait for
-      /*if( 0 ){ return;} //Insert check for DivRecs here, or don't use validHandle below.
-        auto const& divrecHandle = evt.template getValidHandle <std::vector<sim::OpDetDivRec>>(fWavLabel);
-        if(divrecHandle.failedToGet()){
-        return;
-        }*/
+      const std::vector<art::InputTag> G4ModuleLabels = (fG4ModuleLabels.empty()) ?
+        std::vector{fG4ModuleLabel} : fG4ModuleLabels;
```
aihimmel, line 21: This is a little bit dangerous, especially since a user might explicitly specify `G4ModuleLabel` while something upstream specifies `G4ModuleLabels`. I know that's not how it is now, but it would create an invisible failure that would be very hard to track down. Could you add a check which throws a configuration error if both `G4ModuleLabel` and `G4ModuleLabels` are non-empty (or both empty, for that matter)?

aihimmel, line 21: This is a little bit dangerous, especially since a user might explicitly specify `G4ModuleLabel` while something upstream specifies `G4ModuleLabels`. I know that's not how it is now, but it would create an invisible failure that would be very hard to track down. Could you add a check which throws a configuration error if both `G4ModuleLabel` and `G4ModuleLabels` are non-empty (or both empty, for that matter)?

larsim/MCCheater/PhotonBackTracker.tcc
```diff
@@ -17,45 +17,52 @@ namespace cheat{
...
+        auto const& btrHandle = evt.template getValidHandle < std::vector < sim::OpDetBacktrackerRecord > > (G4ModuleLabel);
+        //      if(btrHandle.failedToGet()){
+        /*  mf::LogWarning("PhotonBackTracker") << "failed to get handle to     simb::MCParticle from "
+         *              << fG4ModuleLabel
+         *                          << ", return";*/ //This is now silent as it is expected to    happen every generation run. It is also temporary while we wait for
+        /*if( 0 ){ return;} //Insert check for DivRecs here, or don't use validHandle below.
+          auto const& divrecHandle = evt.template getValidHandle <std::vector<sim::OpDetDivRec>>(fWavLabel);
+          if(divrecHandle.failedToGet()){
+          return;
+          }*/
```
aihimmel: Fine to remove this commented text. It was removed because these noisy but totally harmless warnings confused many users.

larsim/MCCheater/PhotonBackTracker.tcc
```diff
@@ -17,45 +17,52 @@ namespace cheat{
...
+          for(auto& d : dr)
+          {
+          if(!d) continue;
+          priv_od_to_DivRec[od]=*d;//->ref();
+          }
+
+          }
+          }else{throw cet::exception("PhotonBackTracker")<<"find Waveforms and DivRecs from BTRs failed.";}
+        */
+      }
```
aihimmel, line 61: So, this DivRec code was commented out since it was DUNE-specific when this was moved to LArSim. A quick search in the dune repos doesn't show the equivalent showing up anywhere, so probably best to keep it here for now. Maybe add an explanatory comment saying this is DUNE-specific code which hasn't been migrated anywhere better yet?

larsim/MCCheater/PhotonBackTracker.tcc
ikatza: Removed

larsim/MCCheater/PhotonBackTracker.tcc
ikatza: Ok, I added a comment about it.

larsim/MCCheater/PhotonBackTracker.tcc
ikatza: Alex and I had a chat about this offline. For now this approach is good enough, won't cause breaking changes and addresses the problem at hand.

ikatza:
Alex and I had a chat about this PR offline. We believe the best would be to remove the single value label and force users to update their code to use the multi labels parameter; users are most likely not getting the right behaviour currently either way.

However we should have a broader discussion on the topic and allow users to update their code.

We would like to have this PR for an upcoming DUNE-VD production, thus we would like to proceed with this minimal change and postpone the breaking changes.

aihimmel (approved):
I am happy with these proposed changes, though I do not know if I can still formally approve pull requests. @lpaulucc I believe can?

lpaulucc:
@aihimmel Nope, I am not authorized to approve this.

aihimmel:
@ikatza -- do you know who to tag to merge this?

lgarren:
@aihimmel this PR will be discussed at the Monday triage meeting.   And thank you for being proactive!

knoepfel:
trigger build

lgarren:
approve

PR #94, opened 2022-05-19
Make calling geometry service service optional.

yangtj207:
If fDefinedVtxHistRange is set to true, we do not need to get histogram ranges from geometry service. If the gdml file does not follow the larsoft convention, geo->DetHalfWidth() etc. will likely crash. Ideally we should not include analysis code (e.g. filling histograms) in the producer module. But I am keeping the histograms which are useful for debugging. 

Thanks to Robert Hatcher for pointing this problem out.


yangtj207:
trigger build

lgarren:
trigger build

absolution1:
approve

lgarren:
approve

PR #96, opened 2022-07-05
Light simulation updates for dune vertical drift

pgreen135:
This PR contains some minor updates to the semi-analytical light simulation for modelling the lateral PDs in dune vertical drift. There will be a corresponding PR in duneopdet with updated parameter sets (still being finalised). It should not impact any other experiment.


knoepfel:
@pgreen135, is the duneopdet PR ready so that we can trigger the build?

pgreen135:
not yet as the optical library part is still being finalized -- but DUNE-VD isn't included in the CIs yet as far as I'm aware, so I believe the build can be triggered without the corresponding PR to check nothing breaks in any of the other experiments

knoepfel:
trigger build

pgreen135:
as far as I can tell the warnings/failures are all unrelated to this PR, just a lot of reference files that need updating!

absolution1:
approve

lgarren:
approve

PR #97, opened 2022-07-18
Include binomial fluctuations to Correlated method of IonizationScintillation

ikatza:
The main addition are the binomial fluctuations, most relevant for showers.

There are some fixes and do some other pending items.

### DIFF
--- a/larsim/ElectronDrift/SimDriftElectrons_module.cc
+++ b/larsim/ElectronDrift/SimDriftElectrons_module.cc
@@ -89,9 +89,6 @@
 #include <utility>
 #include <vector>
 
-// stuff from wes
-#include "larsim/IonizationScintillation/ISCalcSeparate.h"
-
 namespace detsim {
 
   // Base class for creation of raw signals on wires.
@@ -157,9 +154,6 @@ namespace detsim {
     double fDriftClusterPos[3];
 
     art::ServiceHandle<geo::Geometry const> fGeometry; ///< Handle to the Geometry service
-
-    // IS calculationg
-    larg4::ISCalcSeparate fISAlg;
 
   }; // class SimDriftElectrons
 
--- a/larsim/ElectronDrift/SimDriftElectrons_module.cc
+++ b/larsim/ElectronDrift/SimDriftElectrons_module.cc
@@ -390,7 +390,7 @@ namespace detsim {
                   tpcGeo.PlanePitch(0, 1) * fRecipDriftVel[1]);
       }
 
-      const int nIonizedElectrons = fISAlg.CalcIonAndScint(detProp, energyDeposit).numElectrons;
+      const int nIonizedElectrons = energyDeposit.NumElectrons(); //fISAlg.CalcIonAndScint(detProp, energyDeposit).numElectrons;
       const double lifetimecorrection = TMath::Exp(TDrift / fLifetimeCorr_const);
       const double energy = energyDeposit.Energy();
 
--- a/larsim/ElectronDrift/SimDriftElectrons_module.cc
+++ b/larsim/ElectronDrift/SimDriftElectrons_module.cc
@@ -390,7 +390,7 @@ namespace detsim {
                   tpcGeo.PlanePitch(0, 1) * fRecipDriftVel[1]);
       }
 
-      const int nIonizedElectrons = energyDeposit.NumElectrons(); //fISAlg.CalcIonAndScint(detProp, energyDeposit).numElectrons;
+      const int nIonizedElectrons = energyDeposit.NumElectrons();
       const double lifetimecorrection = TMath::Exp(TDrift / fLifetimeCorr_const);
       const double energy = energyDeposit.Energy();
 
--- a/larsim/ElectronDrift/SimDriftElectrons_module.cc
+++ b/larsim/ElectronDrift/SimDriftElectrons_module.cc
@@ -386,7 +386,7 @@ namespace detsim {
                   tpcGeo.PlanePitch(0, 1) * fRecipDriftVel[1]);
       }
 
-      const int nIonizedElectrons = fISAlg.CalcIonAndScint(detProp, energyDeposit).numElectrons;
+      const int nIonizedElectrons = energyDeposit.NumElectrons(); //fISAlg.CalcIonAndScint(detProp, energyDeposit).numElectrons;
       const double lifetimecorrection = TMath::Exp(TDrift / fLifetimeCorr_const);
       const double energy = energyDeposit.Energy();
 
--- a/larsim/IonizationScintillation/ISCalcCorrelated.cxx
+++ b/larsim/IonizationScintillation/ISCalcCorrelated.cxx
@@ -82,11 +82,12 @@ namespace larg4 {
 
     double ds = edep.StepLength();
     double dEdx = (ds <= 0.0) ? 0.0 : energy_deposit / ds;
+    dEdx = (dEdx < 1.) ? 1. : dEdx;
     double EFieldStep = EFieldAtStep(detProp.Efield(), edep);
     double recomb = 0.;
 
     //calculate recombination survival fraction value inside, otherwise zero
-    if(EFieldStep > 0. && dEdx > 0) {
+    if(EFieldStep > 0.) {
       // calculate recombination survival fraction
       // ...using Modified Box model
       if (fUseModBoxRecomb) {
@@ -97,18 +98,30 @@ namespace larg4 {
       else {
         recomb = fRecombA / (1. + dEdx * fRecombk / EFieldStep);
       }
-      // Guard against unphysical recombination values
-      recomb = (recomb < 0.) ? 0. : recomb;
     }
 
     if(fUseModLarqlRecomb){ //Use corrections from LArQL model
       recomb += EscapingEFraction(dEdx)*FieldCorrection(EFieldStep, dEdx); //Correction for low EF
     }
 
+    // Guard against unphysical recombination values
+    if (recomb < 0.){
+      mf::LogWarning("ISCalcCorrelated")
+        << "Recombination survival fraction is lower than 0.: " << recomb
+        << ", fixing it to 0.";
+      recomb = 0.;
+    }
+    else if (recomb > 1.){
+      mf::LogWarning("ISCalcCorrelated")
+        << "Recombination survival fraction is higher than 1.: " << recomb
+        << ", fixing it to 1.";
+      recomb = 1.;
+    }
+
     // using this recombination, calculate number of ionization electrons
     double num_electrons = (fUseBinomialFlucts) ?
       fBinomialGen.fire(num_ions, recomb) : (num_ions * recomb);
-    num_electrons = (num_electrons > 0.) ? num_electrons : 0.;
+
     // calculate scintillation photons
     double num_photons = (num_quanta - num_electrons) * fScintPreScale;
 
--- a/larsim/IonizationScintillation/ISCalcCorrelated.cxx
+++ b/larsim/IonizationScintillation/ISCalcCorrelated.cxx
@@ -36,7 +36,7 @@ namespace larg4 {
     , fSCE(lar::providerFrom<spacecharge::SpaceChargeService>())
     , fBinomialGen{CLHEP::RandBinomial(Engine)}
   {
-    MF_LOG_INFO("ISCalcCorrelated") << "IonizationAndScintillation/ISCalcCorrelated Initialize." << std::endl;
+    MF_LOG_INFO("ISCalcCorrelated") << "IonizationAndScintillation/ISCalcCorrelated Initialize.";
 
     fScintPreScale = lar::providerFrom<detinfo::LArPropertiesService>()->ScintPreScale();
 
@@ -52,6 +52,7 @@ namespace larg4 {
     fModBoxB = LArG4PropHandle->ModBoxB() / detProp.Density(detProp.Temperature());
     fUseModBoxRecomb = (bool)LArG4PropHandle->UseModBoxRecomb();
     fUseModLarqlRecomb  = (bool)LArG4PropHandle->UseModLarqlRecomb();
+    fUseBinomialFlucts  = (bool)LArG4PropHandle->UseBinomialFlucts();
     fLarqlChi0A = LArG4PropHandle->LarqlChi0A();
     fLarqlChi0B = LArG4PropHandle->LarqlChi0B();
     fLarqlChi0C = LArG4PropHandle->LarqlChi0C();
@@ -111,7 +112,8 @@ namespace larg4 {
     }
 
     // using this recombination, calculate number of ionization electrons
-    double num_electrons = fBinomialGen.fire(num_ions, recomb);
+    double num_electrons = (fUseBinomialFlucts) ?
+      fBinomialGen.fire(num_ions, recomb) : (num_ions * recomb);
     num_electrons = (num_electrons > 0.) ? num_electrons : 0.;
     // calculate scintillation photons
     double num_photons = (num_quanta - num_electrons) * fScintPreScale;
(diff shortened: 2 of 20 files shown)


ikatza:
Marked as draft, want to others to review it first.

knoepfel:
@ikatza, whom would you like to review this?

larsim/IonizationScintillation/ISCalcCorrelated.cxx
```diff
@@ -104,60 +113,28 @@ namespace larg4 {
...
     }
 
-
-   
-
     // using this recombination, calculate number of ionization electrons
-    double const num_electrons = (energy_deposit / fWion) * recomb;
-
+    double num_electrons = fBinomialGen->fire(num_ions, recomb);
+    num_electrons = (num_electrons > 0.) ? num_electrons : 0.;
```
knoepfel: Can a negative number ever be generated from the binomial distribution?  I'm skeptical.

larsim/IonizationScintillation/ISCalcCorrelated.cxx
```diff
@@ -24,24 +24,28 @@
...
   //----------------------------------------------------------------------------
-  ISCalcCorrelated::ISCalcCorrelated(detinfo::DetectorPropertiesData const& detProp)
+  ISCalcCorrelated::ISCalcCorrelated(detinfo::DetectorPropertiesData const& detProp, CLHEP::HepRandomEngine& Engine)
     : fISTPC{*(lar::providerFrom<geo::Geometry>())}
+    , fSCE(lar::providerFrom<spacecharge::SpaceChargeService>())
+    , fBinomialGen(std::make_unique<CLHEP::RandBinomial>(Engine))
   {
-    std::cout << "IonizationAndScintillation/ISCalcCorrelated Initialize." << std::endl;
-    art::ServiceHandle<sim::LArG4Parameters const> LArG4PropHandle;
+    MF_LOG_INFO("ISCalcCorrelated") << "IonizationAndScintillation/ISCalcCorrelated Initialize." << std::endl;
```
knoepfel: Suggest removing `std::endl`--the message logger will automatically include a newline.

larsim/IonizationScintillation/ISCalcCorrelated.cxx
```diff
@@ -104,60 +113,28 @@ namespace larg4 {
...
-    double const num_photons = (Nq - num_electrons) * fScintPreScale;
+    double num_photons = (num_quanta - num_electrons) * fScintPreScale;
 
     MF_LOG_DEBUG("ISCalcCorrelated")
-      << " Electrons produced for " << energy_deposit << " MeV deposited with " << recomb
-      << " recombination: " << num_electrons << std::endl;
-    MF_LOG_DEBUG("ISCalcCorrelated") << "number photons: " << num_photons;
+      << "With " << energy_deposit << " MeV of deposited energy, "
+      << "and a recombination of " << recomb << ", \nthere are "
+      << num_electrons << " electrons, and " << num_photons   <<  " photons.\n";
```
knoepfel: Suggest removing `\n` at end of message for same reason given above.

larsim/IonizationScintillation/ISCalcCorrelated.h
```diff
@@ -19,18 +19,22 @@
...
 namespace spacecharge {
   class SpaceCharge;
 }
 
-namespace detinfo {
-  class LArProperties;
+namespace CLHEP {
+  class HepRandomEngine;
+  class RandBinomial;
 }
```
knoepfel: See comment below.

larsim/IonizationScintillation/ISCalcCorrelated.h
```diff
@@ -39,10 +43,14 @@ namespace larg4 {
                                sim::SimEnergyDeposit const& edep) override;
 
   private:
+    ISTPC fISTPC;
+    const spacecharge::SpaceCharge* fSCE;
+    std::unique_ptr<CLHEP::RandBinomial> fBinomialGen;
```
knoepfel: I applaud your desire to keep dependencies as minimal as possible.  However, in this case, I would explicitly include the header files.  The reasons:

- Anybody who uses this class already has to provide the CLHEP dependencies anyway.
- A `std::unique_ptr<CLHEP::RandBinomial>` data member makes sense if there are cases where no random number is necessary.  However, because this class cannot function without someone providing a random-number engine, the pointer adds more complexity (e.g. an explicit class d'tor) than independence.

The simplest approach is to just own the random-number distribution by value.

larsim/PhotonPropagation/PropagationTimeModel.h
```diff
@@ -105,7 +107,7 @@ class PropagationTimeModel {
 
   // random numbers
   CLHEP::HepRandomEngine& fScintTimeEngine;
-  CLHEP::RandFlat fUniformGen;
+  std::unique_ptr<CLHEP::RandFlat> fUniformGen;
```
knoepfel: See comment above re. using `std::unique_ptr` vs. just owning by value.

knoepfel (commented):
@ikatza, this is a nice, clean PR.  I have a few suggestions for you, but nothing major.  Because this is an interface change, it should be mentioned at a LArSoft coordination meeting.  Do you know if any experiments will require feature branches/PRs?

larsim/IonizationScintillation/ISCalcCorrelated.cxx
ikatza: CLHEP implementation does throw the occasional -1. Maybe for some odd or unphysical combinations, but I do have to guard against it.

larsim/IonizationScintillation/ISCalcCorrelated.cxx
ikatza: Agree

larsim/IonizationScintillation/ISCalcCorrelated.h
ikatza: Fair, I've reverted the commit that made the change

ikatza:
Thanks for your feedback @knoepfel, I've addressed your comments and recommendations.

I've asked @wforeman to review the code, we've messaged on slack.

I added an option to be able to disable the binomial fluctuations. With this change fluctuations will be used by default, which is the normal behaviour. As such there's no need for experiments to create a PR to make use of this, but they should if they want to disable them.

lgarren:
@ikatza @wforeman  Is the review completed to your satisfaction?

wforeman:
Hello, sorry for the delay. Very nice PR. My only reservation is having binomial fluctuations enabled by default. Was a test done to see what this adds to the computation time per event? The expected effect on resolution is quite small (< 1% for 1 MeV of charge dep), which will be dwarfed by natural dE/dx fluctuations. If the impact on speed is significant, perhaps keeping it disabled by default is the better option?

larsim/IonizationScintillation/ISCalcCorrelated.cxx
```diff
@@ -70,7 +81,7 @@ namespace larg4 {
 
     if(EFieldStep > 0) {
       // Guard against spurious values of dE/dx. Note: assumes density of LAr
-      if (dEdx < 1.) dEdx = 1.;
+      //if (dEdx < 1.) dEdx = 1.;
```
wforeman: Are we sure want to remove this? The modified box model turns sharply downward for dE/dx < ~0.8 MeV/cm, and even goes negative < 0.2 MeV/cm. So even a single spuriously low dE/dx could mess things up in unexpected ways. Can we add a check on `recomb` around line 101 instead, to ensure this doesn't go below 0? 
`recomb = (recomb < 0) ? 0. : recomb`

ikatza:
To update on the status of this PR:

I've reached out to Will about how to best guard against unphysical events. Whilst doing some tests I found some issues, these may be related to this PR but I'm not sure yet. I'm investigating and perhaps will tag the solution in here or onto a new PR.

larsim/IonizationScintillation/ISCalcCorrelated.cxx
ikatza: Turns out that yes, best thing is guard against having values of `dEdx` lower than 1.

ikatza:
The last commits fix an issue where the LArQL corrections would overshoot for `dEdx<1.` and `Efield == 0.`, thus relevant for interactions outside the active volume.

I chatted with Franciole Marinho (@ fcmarinho), one of the authors of the LArQL corrections, about adopting this solution. @lpaulucc should vouch the change, as she implemented it in `larsim`.

The increase in running time from using binomial fluctuations is minimal, and in my opinion well worth it to have the binomial fluctuations as default. 

I've run the CI validation tests on SBNDCode, it seems to me the changes are OK. 

The result of the larg4 pds simulation are in:
https://dbweb9.fnal.gov:8443/LarCI/app/ns:sbnd/build_detail/phase_details?build_id=sbnd_ci/9962&platform=Linux%203.10.0-1160.62.1.el7.x86_64&phase=ci_validation&buildtype=slf7%20e20:prof

The full validation test of the larg4 tpc simulation are not working, but I can get the small stats test:
https://dbweb9.fnal.gov:8443/TestCI/app/ns:sbnd/build_detail/phase_details?build_id=sbnd_ci/9968&platform=Linux%203.10.0-1160.62.1.el7.x86_64&phase=ci_validation&buildtype=slf7%20e20:prof
showing that the integrated ADC counts are similar.

lpaulucc (approved):
Modifications look good to me. Thanks, Iker, for fixing the LArQL issue.

knoepfel:
@ikatza, can this PR now be moved out of draft mode?  That is, is it ready for review?

ikatza:
Sorry, I forgot I had marked it as Draft.

knoepfel:
trigger build

knoepfel:
@ikatza, it looks like the ICARUS and MicroBooNE CI/unit tests failed.  Are you able to create feature branches for those experiments to use that accommodate these changes?  Or have you already done that?

ikatza:
Hi @knoepfel, I've opened PR https://github.com/SBNSoftware/icaruscode/pull/429

I'm not sure what's the problem with MicroBooNE, the most I could find is from:
https://dbweb8.fnal.gov:8443/LarCI/app/ns:uboone/storage/docs/2022/08/08/reco_uboone_mcc9_8_driver_overlay_stage1a%23UBUmaqA.out

where WireCell complaints about an empty waveform.

The .err file reports mismatches on the channel noise (?), but I believe those are expected:
```
OneChannelNoise: WARNING: 2 config/data mismatches. #spec=9592, #wave=9595.
```
https://dbweb8.fnal.gov:8443/LarCI/app/ns:uboone/storage/docs/2022/08/08/reco_uboone_mcc9_8_driver_overlay_stage1a%23zonE0nh.err

I'm not familiar with MicroBooNE code so not too sure how to proceed. Do you know of an expert I can query about this?

knoepfel:
@ikatza, I'm not sure the MicroBooNE error is actually related to this PR.  @hgreenlee, can you comment on the error Iker mentions at https://github.com/LArSoft/larsim/pull/97#issuecomment-1215591538?  Is this expected?

larsim/ElectronDrift/SimDriftElectrons_module.cc
```diff
@@ -390,7 +390,7 @@ namespace detsim {
                   tpcGeo.PlanePitch(0, 1) * fRecipDriftVel[1]);
       }
 
-      const int nIonizedElectrons = fISAlg.CalcIonAndScint(detProp, energyDeposit).numElectrons;
+      const int nIonizedElectrons = energyDeposit.NumElectrons();
```
wesketchum, line 387: This is likely an issue for MicroBooNE: I don't think there's a separate IonAndScint module that is performed for uboone, and so it's likely that the number of electrons in the energy deposit that is being fed is -1 and thus empty waveforms into wirecell. This may also be the current issue for ICARUS.

The dirty/do no harm fix is to check if num electrons is -1, and if so use ISAlg to calculate them. The clean/leave things better than you found them fix would be to borrow the code from IonAndScint module that allows for different ISAlgs (not just Separate, as is done here), and add an option for 'None' which can then be used to key in on to use the energyDeposit number of electrons directly. If that's done, default option should probably be 'Separate', and SBND will definitely need to then specify the 'None' option in configs.

wesketchum (changes requested):
See comment on what I think is causing the uboone failures (and what is clearly an issue).

Beyond that, it looks ok ... but I am a little worried about the addition of the randomness in the binomial fluctuations for reproducibility.
* Is the binomial fluctuation RNG seed set someplace?
* Is that engine shared with any other uses that may affect reproducibility?
* Can we verify that we get reproducible results using that seed in jobs that run LArG4 + IonScint, and then again from jobs that run only IonScint from the same LArG4 output?
I'm just paranoid on this, as we've had trouble tracking these things down before. Maybe these checks have already been done though (and if so, great!).

larsim/ElectronDrift/SimDriftElectrons_module.cc
ikatza: Hi Wes, thanks for your comment.

I don't like either of the two solutions.

The clean option is essentially creating a copy of `IonAndScint_module` inside `SimDriftElectrons_module`, which I personally don't feel is a cleaner solution.

The dirty fix won't work if I check `numElectrons == -1`, the default constructor of [`SimEnergyDeposit`](https://github.com/LArSoft/lardataobj/blob/9cc938ecdc5940b1dd7840bd3a95ab6ae51caae4/lardataobj/Simulation/SimEnergyDeposit.h#L65) sets `numElectrons = 0`. Having 0 electrons is not extraordinary. Thus the dirty fix would require to modify `lardataobj` to set `-1` as default. Perhaps checking a combination of various conditions (checking that electrons and photons are == 0, but other variables aren't) could be a sign that the number of electrons hasn't been calculated already.

I'll attempt to see if I can figure out a dirty solution along these lines.

ikatza: Well I came up with a hacky solution to test `energyDeposit` on `icaruscode` and turns out  `ElectronDrift/SimDriftElectrons_module.cc` is not run on `icaruscode`, at least not on the fcl used on the test `icarus_ci_single_g4_quick_test_icaruscode.fcl` nor on `standard_g4_icarus.fcl`.

I suspect MicroBooNE doesn't run it either since it shares the same name of the process in the TimeTracker table

```
simulate:ionization:larsim/ElectronDrift/ShiftEdepSCE
``` 

By the way that the last MicroBoone test of the G4 stage ran OK:
https://dbweb8.fnal.gov:8443/LarCI/app/ns:uboone/storage/docs/2022/08/08/wirecell_g4_uboone%23oJyVHdX.out

ikatza:
In my last commit , I've renamed the seed to hopefully address @wesketchum comments about the RNG seed. To answer his observations:

> - Is the binomial fluctuation RNG seed set someplace?

Yes, in `IonAndScint_module.cc`


> - Is that engine shared with any other uses that may affect reproducibility?

Not simultaneously, it's also used for `ISCalcNEST` but the `ISCalcAlg`'s are exclusive. Thus effectevely the engine won't be shared.


> - Can we verify that we get reproducible results using that seed in jobs that run LArG4 + IonScint, and then again from jobs that run only IonScint from the same LArG4 output?
I'm just paranoid on this, as we've had trouble tracking these things down before. Maybe these checks have already been done though (and if so, great!).

I didn't test if the numbers are reproducible, rather I assume that it will work fine as it follows the same procedure that other modules. I don't quite follow what the two verification scenarios are, but I suspect that one of them is only relevant in the `LegacyLArG4`, which this PR does not modify.

larsim/ElectronDrift/SimDriftElectrons_module.cc
ikatza: To summarise, I don't believe this is the reason for the failed MicroBoone test. It's definitively not the reason for the ICARUS failure, which I addressed in https://github.com/SBNSoftware/icaruscode/pull/429

knoepfel:
trigger build

knoepfel:
@wesketchum, are you satisfied with @ikatza's responses?

lgarren:
@wesketchum do you approve?

wesketchum:
yes, if the tests are passing now I approve.

wesketchum (approved):

lgarren:
approve

PR #98, opened 2022-09-02
Add module and tool for photon fast simulation

Healthborn:
Add module and tool for photon fast simulation

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -61,6 +61,9 @@ find_package(lardataalg REQUIRED EXPORT)
 find_package(lardataobj REQUIRED EXPORT)
 find_package(larevt REQUIRED EXPORT)
 
+find_package(TensorFlow 2.6.0 QUIET EXPORT)
+find_package (Eigen3 REQUIRED NO_MODULE)
+
 # Wes put this here to use TRACE for debugging...
 #find_ups_product( TRACE )
 
--- a/Modules/FindTensorFlow.cmake
+++ b/Modules/FindTensorFlow.cmake
@@ -1,0 +1,110 @@
+#[============================================================[.rst:
+FindTensorFlow
+============
+#]============================================================]
+if (UNIX)
+  set(_ftf_m m)
+  if (APPLE)
+    set(_ftf_rt)
+  else()
+    set(_ftf_rt rt)
+  endif()
+else()
+  set(_ftf_m)
+  set(_ftf_rt)
+endif()
+set(_ftf_libs "_framework" "" "_cc")
+set(_ftf_transitive_deps TensorFlow::framework)
+set(_ftf_transitive_deps_cc TensorFlow::framework)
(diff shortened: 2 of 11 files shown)


lgarren:
@Healthborn LArSoft use of tensorflow is modularized.  That means it is only meant to be used through larrecodnn.

Healthborn:
> 

@lgarren, I know there is a product larsimdnn, can I commit to that product? But it seems larsimdnn is out of date.

knoepfel:
@Healthborn, thanks for the feedback and the PR.  The SciSoft team will discuss this and propose a way forward.

knoepfel:
@Healthborn, we think this PR belongs in https://github.com/LArSoft/larsimdnn.  However, we need to update that repository to support cetmodules and we will supply the `FindTensorFlow.cmake` module in a different way.  Once we have those things in place, we will ask you to close this PR and create a new PR in `larsimdnn`.

Healthborn:
Thank you @knoepfel!

lgarren:
@Healthborn We believe that the head of develop of larsimdnn is now ready for you to use.  You will need to build both larfinder (new) and larsimdnn.   Please contact the SciSoft team if you have any problems.

Healthborn:
@lgarren I'm sorry, I don't understand what do you mean by "You will need to build both larfinder (new) and larsimdnn". Can I check in my code to only larsimdnn?

lgarren:
@Healthborn  You should only commit your code to larsimdnn.  However, the tensorflow find module is now in the new larfinder package.  This allows the same find module to be used by both larsimdnn and larrecodnn.  Since larfinder is not yet part of a release, you will need to build both larsimdnn and larfinder.

Healthborn:
@lgarren  Does that mean: I need not add \Modules\FindTensorFlow.cmake to my checkin, and I need only check in my code and update the CMakeLists.txt?

Healthborn:
@lgarren When I tried to check out larfinder, I got the following error:
mrb g larfinder
INFO: Unable to verify write access for larfinder, falling back to read-only.
ERROR: Unable to verify read access to remote http://cdcvs.fnal.gov/projects/larfinder.

lgarren:
@Healthborn, I suggest sending further quesxtions to scisoft-team@fnal.gov.  You'll need a newer mrb, which is not yet available on larsoft cvmfs.   You can just `git clone git@github.com:LArSoft/larfinder.git` and then `mrb uc`

Healthborn:
@lgarren I tried to run: git clone git@github.com:LArSoft/larfinder.git, and got the following error:
Cloning into 'larfinder'...
Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.

lgarren:
Oh, right.  You need the read-only clone: `https://github.com/LArSoft/larfinder.git`

Healthborn:
It works! Many thanks!

PR #118, opened 2023-06-02
New orientation for the PDs added for new DUNE-VD v5 geometry

Li-Jiaoyang97:
Modifications being made are:
- added a new orientation (=2) for the new lateral PDs (fixed at z-direcrtion) for DUNE-VD v5 geometry. 
- changed the code to be adapted with the new orientation
- removed the code where applying field cage transparency factor because of the implementation of hybrid model, which covers the region outside the field cage.

### DIFF
--- a/larsim/PhotonPropagation/SemiAnalyticalModel.cxx
+++ b/larsim/PhotonPropagation/SemiAnalyticalModel.cxx
@@ -194,7 +194,7 @@ namespace phot {
     geo::Vector_t const relative = ScintPoint - opDet.center;
     const double distance = relative.R();
     double cosine;
-    if (opDet.orientation == 2) 
+    if (opDet.orientation == 2)
       cosine = std::abs(relative.Z()) / distance;
     else if (opDet.orientation == 1)
       cosine = std::abs(relative.Y()) / distance;
--- a/larsim/PhotonPropagation/SemiAnalyticalModel.cxx
+++ b/larsim/PhotonPropagation/SemiAnalyticalModel.cxx
@@ -194,7 +194,9 @@ namespace phot {
     geo::Vector_t const relative = ScintPoint - opDet.center;
     const double distance = relative.R();
     double cosine;
-    if (opDet.orientation == 1)
+    if (opDet.orientation == 2)
+      cosine = std::abs(relative.Z()) / distance;
+    else if (opDet.orientation == 1)
       cosine = std::abs(relative.Y()) / distance;
     else
       cosine = std::abs(relative.X()) / distance;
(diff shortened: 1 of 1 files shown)


knoepfel:
@Li-Jiaoyang97, thanks for the PR.  Do you know if any framework job (for DUNE or any other experiment) has previously relied on an orientation value of "2"?  These code changes would subtly affect such jobs if any exist.

Li-Jiaoyang97:
Hi @knoepfel, as far as I am aware, there is no experiment has previously relied on an orientation value of "2". The DUNE-VD newest geometry **will** rely on this orientation value. I can ask @pgreen135 to confirm. Thank you both so much!

knoepfel:
trigger build

lgarren:
@LArSoft/level-2-managers  Please take a close look at this PR.  We are concerned that there may be subtle unexpected differences if any experiment code presently uses a value of 2 for orientation.

Li-Jiaoyang97:
Hi @lgarren, I will check and confirm with SBND and DUNE-HD. Thanks!

pgreen135:
@lgarren I've asked Jiaoyang to check running the code in SBND and DUNE-HD to verify there are no unintended changes.  Ideally a better method would be used to identify the optical detector orientations, however is not available currently in the geometry service tools (there was a discussion about adding this a while ago, but I'm not sure it ever happened; and is a larger project). No other experiment uses the orientations "1" and "2" as they are specific to the photodetectors on the sides of the detector proposed for DUNE-VD, so there shouldn't be an issue; but please hold off merging until we confirm. Thanks!

lgarren:
Thanks @pgreen135   You'll have seen the email I just sent to interested parties.

lgarren:
@pgreen135  We suspected that the default is zero unless set.  Can you verify that?

pgreen135:
@lgarren yes the default is zero, which corresponds to the standard photo-detectors at the anode as in DUNE-HD, SBND, uBooNE, etc.

lgarren:
@Li-Jiaoyang97 @pgreen135 what is the status?  We're still waiting for checks.

Li-Jiaoyang97:
Hi @lgarren , 
Unfortunately I have not finished the validation yet. My plan was to finish it last week. But due to the 2.5-day notice of my talk in New Perspective, my plan was completely interrupted. I will ping you when I finish the validation. 
Thanks!
Jiaoyang

erica-snider:
@Li-Jiaoyang97 and others. I think the question about the orientation was slightly mis-stated. The concern here is less about whether the orientation value "2" is used elsewhere -- I'm willing to believe that this is a newly introduced value -- but rather the robustness of the OpticalDetector "width" vs "length" test that is used to set it. This seems to rest on a fragile set of assumptions about the shape of optical detectors, and will lead to unintended results if those assumptions turn out to be wrong in some case. It might also be difficult to locate such a problem unless this requirement is well documented, which it is not. 

The longer term solution is clearly to provide a direct answer to the orientation question in the geometry services. In the short term, we need to know whether the "length" vs "width" test returns the correct answer for all existing optical detectors.

Erica

Li-Jiaoyang97:
Hi @erica-snider and @lgarren and all. I have done the test with SBND, DUNE-HD and ProtoDUNE-HD geometries (these are all I could think that are using the semi-analytical model for the fast optical simulation and with XARAPUCAS). 
All the existing XARPUAS will return the default values which are zero. I have also check the truth level of number of photons before and after my modifications, they are the same. I have been tested with `larsim` with tag `LARSOFT_SUITE_v09_75_00` for all three different experiments which is what I used for the implementation. Please let me know if there is any other validations you want me to do. Thank you all very much! I really appreciate the patience.

lgarren:
@Li-Jiaoyang97 thanks for all your hard work!

lgarren:
trigger build

lgarren:
approve

PR #119, opened 2023-06-20
Adding alpha quenching to the Correlated IonAndScint method

lpaulucc:
The DUNE low energy group requested a separate treatment for the alpha particle when calculating the ionization and scintillation obtained from it. Alpha particles produce a reduced scintillation signal compared to electrons of the same energy, an effect known as quenching due to high excitation density. To address this, the modification of the charge yield added by LArQL (escaping electrons and electric field corrections) was removed for the alpha particle. In this way, alpha LY will not depend on the electric field (it has a small dependence on the EF but the _opposite_ one that is currently being calculated by LArQL). A fhicl parameter, the quenching factor, QAlpha, was introduced with default value of 0.72 following Doke et al. 2002. With this value, the alpha LY is ~36000 ph/MeV. This factor affects both light and charge.
References: A Hitachi, A Yunoki, T Doke, T Takahashi - PRA 35, 3956 (1987); Doke et al. Jpn. J. Appl. Phys. v. 41 (2002) pt. 1, 3A
A small correction to the method was also added: binomial sample on the number of electrons generated is only called if the deposited energy is higher than the ionization threshold.

### DIFF
--- a/larsim/IonizationScintillation/ISCalcCorrelated.cxx
+++ b/larsim/IonizationScintillation/ISCalcCorrelated.cxx
@@ -59,8 +59,8 @@ namespace larg4 {
     fLarqlChi0C = LArG4PropHandle->LarqlChi0C();
     fLarqlChi0D = LArG4PropHandle->LarqlChi0D();
     fLarqlAlpha = LArG4PropHandle->LarqlAlpha();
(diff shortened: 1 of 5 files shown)


lgarren:
trigger build

absolution1:
approve

lgarren:
approve

PR #121, opened 2023-08-07
Flexible path for radiogen files

lpaulucc:
Radiological files needed for the next FD dune production are in StashCache but the RadioGen module looked for files with a hard coded path. A new fhicl parameter was added to the module so that it is flexible enough to find files with a change in the configuration block and not the module itself.

### DIFF
--- a/larsim/EventGenerator/RadioGen_module.cc
+++ b/larsim/EventGenerator/RadioGen_module.cc
@@ -27,6 +27,9 @@
 //             Ni59 Calibration sources. This is another "hacky"
 //             fix to something that really deserves a more
 //             elegant and comprehensive solution.
+//
+//           Aug 2023 L Paulucci
+//             Adding a fhicl parameter for generic path to radiogen files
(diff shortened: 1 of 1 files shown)


knoepfel:
@lpaulucc, this looks like a good improvement.  Now that the file-path is configurable, at the very least, you should add a check to make sure that the last character is a forward slash `'/'`.

lpaulucc:
Thank you @knoepfel. I have added the check of the last character on the path and included an exception if it is not a forward slash. Please let me know if I didn't use the appropriate case for that.

knoepfel:
trigger build

knoepfel:
@lpaulucc, the `'/'` check looks good.  We'll see how the CI reacts to the change.

lgarren:
@LArSoft/level-2-managers please review and approve or comment

tomjunk:
Approved

Looks fine.  Thanks!

lgarren:
approve

PR #125, opened 2023-10-31
Adding An Ellipsoidal Model for Recombination

jzennamo:
This PR adds a simulation of an ellipsoidal model for recombination which creates an angular dependence for the recombination model. This is based on this paper https://iopscience.iop.org/article/10.1088/1748-0221/12/12/P12002/pdf and developed further by Gray Putnam (UChicago).

### DIFF
--- a/larsim/IonizationScintillation/ISCalcCorrelated.cxx
+++ b/larsim/IonizationScintillation/ISCalcCorrelated.cxx
@@ -117,8 +117,7 @@ namespace larg4 {
         }
         else {
           double B_ellips = fEllipsModBoxB * dEdx /
-                            (EFieldStep * sqrt(pow(std::sin(phi), 2) +
-                                               (pow(std::cos(phi), 2) / pow(fEllipsModBoxR, 2))));
+                            (EFieldStep * std::hypot(std::sin(phi),  std::cos(phi) / fEllipsModBoxR));
 
           recomb = std::log(fEllipsModBoxA + B_ellips) / B_ellips;
         }
--- a/larsim/IonizationScintillation/ISCalcCorrelated.cxx
+++ b/larsim/IonizationScintillation/ISCalcCorrelated.cxx
@@ -79,94 +79,98 @@ namespace larg4 {
                                                sim::SimEnergyDeposit const& edep)
   {
 
-      double const energy_deposit = edep.Energy();
-      
-      // calculate total quanta (ions + excitons)
-      double num_ions = 0.0; //check if the deposited energy is above ionization threshold
-      if (energy_deposit >= fWion) num_ions = energy_deposit / fWion;
-      double num_quanta = energy_deposit / fWph;
-      
-      double ds = edep.StepLength();
-      double dEdx = (ds <= 0.0) ? 0.0 : energy_deposit / ds;
-      dEdx = (dEdx < 1.) ? 1. : dEdx;
-      double EFieldStep = EFieldAtStep(detProp.Efield(), edep);
-      double recomb = 0., num_electrons = 0.;
-      
-      //calculate recombination survival fraction value inside, otherwise zero
-      if (EFieldStep > 0.) {
-	// calculate recombination survival fraction
-	// ...using Modified Box model
-	if (fUseModBoxRecomb) {
-	  double Xi = fModBoxB * dEdx / EFieldStep;
-	  recomb = std::log(fModBoxA + Xi) / Xi;
-      }
-	else if (fUseEllipsModBoxRecomb){
+    double const energy_deposit = edep.Energy();
 
-	  double phi = std::acos(abs(edep.StartX() - edep.EndX()) / sqrt(pow(edep.StartX() - edep.EndX(), 2) + pow(edep.StartY() - edep.EndY(), 2) + pow(edep.StartZ() - edep.EndZ(), 2)));
-	  
-	  if(phi > std::atan(1)*2){
-	    double temp_phi = phi;
-	    phi = std::atan(1)*4-temp_phi;
-	  }
-	  
-	  if(phi != phi){
-	    double Xi = fModBoxB * dEdx / EFieldStep;
-	    recomb = std::log(fModBoxA + Xi) / Xi;
-	  }	  
-	  else{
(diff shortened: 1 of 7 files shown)


larsim/IonizationScintillation/ISCalcCorrelated.cxx
```diff
@@ -96,6 +100,30 @@ namespace larg4 {
...
+      else if (fUseEllipsModBoxRecomb) {
+
+        double phi =
+          std::acos(abs(edep.StartX() - edep.EndX()) /
+                    sqrt(pow(edep.StartX() - edep.EndX(), 2) + pow(edep.StartY() - edep.EndY(), 2) +
+                         pow(edep.StartZ() - edep.EndZ(), 2)));
+
+        if (phi > std::atan(1) * 2) {
+          double temp_phi = phi;
+          phi = std::atan(1) * 4 - temp_phi;
```
knoepfel:
```suggestion
          phi = std::atan(1) * 4 - phi;
```

larsim/IonizationScintillation/ISCalcCorrelated.cxx
```diff
@@ -96,6 +100,30 @@ namespace larg4 {
...
+        }
+
+        if (phi != phi) {
+          double Xi = fModBoxB * dEdx / EFieldStep;
+          recomb = std::log(fModBoxA + Xi) / Xi;
+        }
+        else {
+          double B_ellips = fEllipsModBoxB * dEdx /
+                            (EFieldStep * sqrt(pow(std::sin(phi), 2) +
+                                               (pow(std::cos(phi), 2) / pow(fEllipsModBoxR, 2))));
```
knoepfel:
```suggestion
                            (EFieldStep * std::hypot(std::sin(phi),  std::cos(phi) / fEllipsModBoxR));
```

larsim/IonizationScintillation/ISCalcCorrelated.cxx
```diff
@@ -96,6 +100,30 @@ namespace larg4 {
         double Xi = fModBoxB * dEdx / EFieldStep;
         recomb = std::log(fModBoxA + Xi) / Xi;
       }
+      else if (fUseEllipsModBoxRecomb) {
+
+        double phi =
+          std::acos(abs(edep.StartX() - edep.EndX()) /
+                    sqrt(pow(edep.StartX() - edep.EndX(), 2) + pow(edep.StartY() - edep.EndY(), 2) +
+                         pow(edep.StartZ() - edep.EndZ(), 2)));
```
knoepfel:
```suggestion
        double phi = std::acos(abs(edep.StartX() - edep.EndX()) / edep.StepLength());
```

knoepfel (commented):
@jzennamo, thanks for the PR.  In general, the changes look fine.  I've suggested some adjustments below for your consideration.

larsim/IonizationScintillation/ISCalcCorrelated.cxx
jzennamo: Awsome! I had no idea this existed!

lgarren:
@jzennamo Are you having a problem running format-code?

jzennamo:
Hi @lgarren, I think I succeeded in running it and I pushed the changes it made

lgarren:
@jzennamo I see that the code checks have just passed.  Thank you

lgarren:
trigger build

lgarren:
@jzennamo please fix the c14 build problems.  See https://github.com/LArSoft/larsim/pull/125#issuecomment-1796306080

jzennamo:
@lgarren I think my above will fix this issue

jzennamo:
also, I fixed more code formatting

jzennamo:
@lgarren OK, I setup this c14:debug build on my end and solved the issue in the above push, you should be good to trigger the build again

lgarren:
trigger build

tomjunk:
A couple of comments:

in simulationservices.fcl, it appears as if the default value of UseModLarqlRecomb has changed from false to true.  The ellipsoidal recomb is false by default -- is the intention not to change the default behavior?

std::atan(1) is called repeatedly.  One can use TMath::Pi(), TMath::TwoPi(), TMath::PiOver2() to not do the calculation.  Or one ca cache the result.

Perhaps most seriously, the calculation of phi appears to assume that the electric field E points along the X direction.  One should get the E field direction from the geometry at the very least, and of cours space charge will point it in slightly different directions.

Is it possible to have zero-length steps?  Dividing by the step length might give NaN.  Might be good to put in some defense against zero-length steps anyway.

jzennamo:
Oops! The change to `simulationservices.fcl` was a mistake, I'll revert that. 

I will move to using `TMath::PiOver2()` instead of `std::atan(1)`.

zero-length steps are treated separately because of the dEdx correction, but in addition that is why there is this section of the code:
```
if (phi != phi) {
  double Xi = fModBoxB * dEdx / EFieldStep;
  recomb = std::log(fModBoxA + Xi) / Xi;
}
 ```

For the electric field direction, I think I would profit from an in person conversation of what we can do for that, I'll come looking for you

tomjunk:
As per our conversation in person, an example of querying the geometry to get the drift direction is 

https://github.com/LArSoft/larsim/blob/6c2fee13cf230ff36094025d8b3de32aad3b301a/larsim/ElectronDrift/SimDriftElectrons_module.cc#L312

Space charge may not be as hard as I thought.  Here's the correction to the electron drift:

https://github.com/LArSoft/larsim/blob/6c2fee13cf230ff36094025d8b3de32aad3b301a/larsim/ElectronDrift/SimDriftElectrons_module.cc#L353

Space charge maps are stored as distortions for electrons that drift over long distances. What you want however is the local E field distortion.  It may be possible to calculate a numerical derivative of the space charge offset with respect to the true position to get the E field distortion.  As mentioned before, this is icing on the cake and may be a second-order effect.

I also prefer

https://en.cppreference.com/w/cpp/numeric/math/isnan

to checking if (phi != phi) because a compiler might just decide to optimize that away.  But if it's part of the standard to say that NaN != NaN and we can rely on that, I suppose it's okay, but std::isnan I think is more clear.  Or you can check the denominator in the phi calc for zero.

jzennamo:
@tomjunk OK! I think I have added a function to measure the angle of the GEANT step relative to the generalized electric field, please take a look at the new `AngleToEFieldAtStep` function.

tomjunk:
Approved

 Thanks, Joe!

lgarren:
approve

PR #126, opened 2023-11-09
Corrected for SCE convention and generalized

jzennamo:
This generalizes the `EFieldAtStep` function and corrects a typo I had in the convention for the direction of SCE electric field offsets. 

What looks like a bug in the correlated ionization and scintillation simulation seems to have the space-charge E-field distortion added incorrectly to the E field.

This PR changes default behavior in DUNE's Monte Carlo simulation, and should be reviewed by simulation experts.

The PR has been tested with ICARUS but it affects any detector with an E field that points in any direction other than +X.

### DIFF
--- a/larsim/IonizationScintillation/ISCalcCorrelated.cxx
+++ b/larsim/IonizationScintillation/ISCalcCorrelated.cxx
@@ -166,11 +166,52 @@ namespace larg4 {
     // electric field outside active volume set to zero
     if (!fISTPC.isScintInActiveVolume(edep.MidPoint())) return 0.;
 
-    // electric field inside active volume
-    if (!fSCE->EnableSimEfieldSCE()) return efield;
+    TVector3 elecvec;
 
-    auto const eFieldOffsets = fSCE->GetEfieldOffsets(edep.MidPoint());
-    return efield * std::hypot(1 + eFieldOffsets.X(), eFieldOffsets.Y(), eFieldOffsets.Z());
+    art::ServiceHandle<geo::Geometry const> fGeometry;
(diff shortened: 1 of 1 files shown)


lgarren:
trigger build

lgarren:
trigger build

lgarren:
@jzennamo  are you able to address these concerns from Mike Mooney?

Please see the previous concern I had about how larsim is handling SCE offsets:

    - https://cdcvs.fnal.gov/redmine/issues/28099
    - https://indico.fnal.gov/event/59975/contributions/268145/attachments/167532/223567/LArSoft_CoordinationMeeting_23_05_30.pdf

This concerns spatial SCE offsets, but any possible redesign to the SpaceCharge service could impact E field distortion offsets as well.

Agreement needs to be reached about how to proceed.  Basically, as of right now, larsim is internally inconsistent.  Once that is addressed, I think it will be straightforward to include the desired impact from SCE in this addition to the ionization/scintillation simulation.

If you would like to avoid any problems in the meantime for the PR in question, I would remove the SCE dependence in Joseph's addition.  Anything we do right now with this will likely be changed, and I'd rather not have to make two sets of changes.  Also, as Joseph demonstrated to me, it is a very, very small effect.

jzennamo:
Hi @lgarren, thanks for bringing this up! This applies specifically to the position offsets. To validate my handling of this convention, I checked the magnitude of the electric field on either side of the cathode to confirm it was correctly handling the sign. I will be happy to update my handling once we have come up with a new convention.

lgarren:
After discussion offline, this PR will be approved and merged.

lgarren:
approve

PR #128, opened 2023-12-12
Store original track id for sim energy deposits

bear-is-asleep:


knoepfel:
trigger build

lgarren:
@LArSoft/level-2-managers please approve or comment!

tomjunk:
Approved.

This fills in a value for an argument in the constructor which defaults to zero which presumably is worse than an actual value for the original track ID.  I only looked at the commits and the DUNE tests which are fine.

lgarren:
approve

PR #129, opened 2024-01-19
Bug Fix; Adding a check that the TPCID is valid in ISCalcCorrelated 

jzennamo:
DUNE-VD informed me that they use `ISCalcCorrelated` to simulate light outside the TPC volume, this was a use case I did not expect. This causes `ISCalcCorrelated` to crash because there was no check that the TPCID is valid. This PR adds this check and enables the DUNE-VD simulation to work.


tomjunk:
Approved.

knoepfel:
trigger build

lgarren:
approve

PR #130, opened 2024-01-24
Handle extremely large times that have undefined behaviour when being converted from double to int

absolution1:
SimPhotonsLite class stores time as an int (its used as the key in a map)
It's possible that long lived Ar isotopes decay many hours after an event's simulation time.  This decay time (as a double/float) is larger than the maximum value of an int.  Converting these extreme values to an int wraps the time to the most extreme negative int.

This PR enforces that the large time becomes a max int.

### DIFF
--- a/larsim/PhotonPropagation/PDFastSimPVS_module.cc
+++ b/larsim/PhotonPropagation/PDFastSimPVS_module.cc
@@ -261,7 +261,7 @@ namespace phot {
                 // calculate the time at which each photon is seen
                 double dtime = edepi.StartT() + fScintTime->fastScintTime();
                 if (fIncludePropTime) dtime += transport_time[i];
-                int time = static_cast<int>(std::round(dtime));
+                int time(dtime > static_cast<double>(std::numeric_limits<int>::max()) ? std::numeric_limits<int>::max() : static_cast<int>(std::round(dtime)));
                 if (Reflected)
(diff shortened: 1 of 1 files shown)


absolution1:
trigger build

knoepfel:
The logic looks good @absolution1.  I might encourage though defining a function in an anonymous namespace in that file and just using where necessary:

```c++
namespace {
  int constrain_to_int_max(double d)
  {
      constexpr auto int_max_as_double = static_cast<double>(std::numeric_limits<int>::max());
      return static_cast<int>(std::min(int_max_as_double, std::round(d)));
  }
}
```
and then later on:

```diff
- int time = static_cast<int>(std::round(dtime));
+ int time = constrain_to_int_max(dtime);
```

absolution1:
Hi @knoepfel 
Thanks for checking in on this.
I think the code suggestion is good, but it would only be reducing one bit of boilerplating in a larger chunk of boilerplate code.  At a glance, it looks like a much more significant amount of this section of code could be anonymous namespace function'd away.
I would like to suggest that the bug fix go in as is (and maintain the level of boilerplate), with a separate task to functionise more of this code.

What do you think?

knoepfel:
Approve.

That's fine @absolution1.  Will clean things up another time.

knoepfel:
approve

PR #132, opened 2024-02-26
Fix geometry service call that expects an exception it is never going to get

AlexWilkinsonnn:
I tried to use the electron drift module to make simchannels from simenergydeposits and found I would get an error complaining about an invalid cryostatID. It is caused by code calling `fGeom->PositionTo{Cryostat,TPC}ID(pos)` expecting an exception if the position is outside the necessary volume. I suppose this was true at some point in the geometry service's past but is no longer true. Instead we need to check if this method returns an invalid ID to know if position is not in a TPC or Cyrostat boundary.

I  found the same code in MCRecoEdep.cxx where I fixed it there also.

### DIFF
--- a/larsim/ElectronDrift/SimDriftElectrons_module.cc
+++ b/larsim/ElectronDrift/SimDriftElectrons_module.cc
@@ -273,24 +273,18 @@ namespace detsim {
       // cryostat and tpc. If somehow the step is outside a tpc
       // (e.g., cosmic rays in rock) just move on to the next one.
       unsigned int cryostat = 0;
-      try {
-        cryostat = fGeometry->PositionToCryostatID(mp).Cryostat;
-      }
-      catch (cet::exception& e) {
+      cryostat = fGeometry->PositionToCryostatID(mp).Cryostat;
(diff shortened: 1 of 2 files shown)


absolution1:
Looks like a good change to me

absolution1:
approve

knoepfel:
Thanks for the PR, @AlexWilkinsonnn.  There's a compilation failure, likely due to a copy-paste error (`MCRecoEdep.cxx` should be using `geom` not `fGeometry`).

AlexWilkinsonnn:
Argh sorry, I checked the first change built then got lazy. Thanks for pointing out the mistake @knoepfel !

lgarren:
trigger build

lgarren:
@AlexWilkinsonnn please check the build failure.   Either a missing header or another copy-paste problem.

AlexWilkinsonnn:
Yes another error with copy pasting :disappointed:

lgarren:
trigger build

lgarren:
approve

PR #135, opened 2024-03-27
Add an automatic voxelisation option to PhotonVisibilityService

bwynneHEP:
Added an automatic voxelisation option, fitting exactly to the TPC boundaries and then optimising for the cryostat

Also added the option to save lists of voxel IDs within the TPC and outside, to allow efficient generation of photon library.

NB, this seems to require a rebuild of larana, specificially due to
https://github.com/LArSoft/larana/blob/b6abb28a9515bb5a8193e6aa2fd131a2ee0ecd8e/larana/OpticalDetector/SimPhotonCounter_module.cc#L192
This seems like a problem with the build system, but maybe I'm missing something obvious.

### DIFF
--- a/larsim/PhotonPropagation/PhotonVisibilityService.cc
+++ b/larsim/PhotonPropagation/PhotonVisibilityService.cc
@@ -315,14 +315,13 @@ namespace phot {
 
       // Find the TPC volume, following example from
       // https://github.com/LArSoft/larsim/blob/05378155a2a07551fa5757d0449ffb30476d5cf9/larsim/LegacyLArG4/OpFastScintillation.cxx#L2149-L2165
-      geo::BoxBoundedGeo cryoVolume{ geom->Cryostat().BoundingBox() };
-      geo::BoxBoundedGeo tpcVolume{ geom->Cryostat().TPC(0).ActiveBoundingBox() };
-      for ( geo::CryostatGeo const& cryo : geom->Iterate<geo::CryostatGeo>() ) {
+      geo::BoxBoundedGeo cryoVolume{geom->Cryostat().BoundingBox()};
+      geo::BoxBoundedGeo tpcVolume{geom->Cryostat().TPC(0).ActiveBoundingBox()};
+      for (geo::CryostatGeo const& cryo : geom->Iterate<geo::CryostatGeo>()) {
 
-        cryoVolume.ExtendToInclude( cryo.BoundingBox() );
-        for ( geo::TPCGeo const& TPC : cryo.IterateTPCs() )
-        {
-          tpcVolume.ExtendToInclude( TPC.ActiveBoundingBox() );
+        cryoVolume.ExtendToInclude(cryo.BoundingBox());
+        for (geo::TPCGeo const& TPC : cryo.IterateTPCs()) {
+          tpcVolume.ExtendToInclude(TPC.ActiveBoundingBox());
         }
       }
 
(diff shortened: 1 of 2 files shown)


knoepfel:
@bwynneHEP, thanks for the PR.  We are trying to understand the motivation for these changes--could you give us some background information?  Presenting this at a LArSoft Coordination Meeting could be helpful.

Regarding the `larana` build issue--a rebuild is required of any downstream code that includes this service as a header and/or links against the corresponding library.

bwynneHEP:
@knoepfel the motivation is simply convenience. The context is generating samples for the "photon library," but perhaps this has more general application.

We want a voxel grid that covers the cryostat efficiently, with boundaries that align to the TPC so that it is easy to label a voxel as inside or outside the TPC. All these dimensions can be looked up, and a voxelisation calculated and added to a configuration file, but it's far quicker and more reliable to just calculate it directly in the PVS from the existing geometry description. Also this includes a simple optimisation step that would otherwise add complexity to the apparently back-of-an-envelope guesswork used before.

The output lists of voxel indices inside or outside the TPC then simplifies sorting jobs and outputs for large-scale production of samples.

bwynneHEP:
I should also note that all this behaviour defaults to off - there is no modification to existing workflows. Any effects would only be felt at configuration time, if the behaviour is enabled.

knoepfel:
@bwynneHEP, sorry for the delay, and thanks for the response.  The idea seems reasonable.  I will send some implementation suggestions later this week.

larsim/PhotonPropagation/PhotonVisibilityService.cc
```diff
@@ -306,7 +313,85 @@ namespace phot {
...
+        std::string logString = "Voxels within TPC: " + std::to_string(tpcVoxels.size());
+        if (fSaveTPCVoxels != "") logString += " saved to file\n\t" + fSaveTPCVoxels;
+        logString += "\nVoxels outside TPC: " + std::to_string(otherVoxels.size());
+        if (fSaveOtherVoxels != "") logString += " saved to file\n\t" + fSaveOtherVoxels;
+        mf::LogInfo("PhotonVisibilityService") << logString << std::endl;
+
+        if (fSaveTPCVoxels != "") {
+          std::ofstream outputFile = std::ofstream(fSaveTPCVoxels);
+          for (auto voxel : tpcVoxels)
+            outputFile << std::to_string(voxel) + "\n";
```
knoepfel: No need to convert voxel to string before streaming to the output file.

```suggestion
            outputFile << voxel << '\n';
```

larsim/PhotonPropagation/PhotonVisibilityService.cc
```diff
@@ -306,7 +313,85 @@ namespace phot {
...
+        }
+        if (fSaveOtherVoxels != "") {
+          std::ofstream outputFile = std::ofstream(fSaveOtherVoxels);
+          for (auto voxel : otherVoxels)
+            outputFile << std::to_string(voxel) + "\n";
```
knoepfel: See above comment.

larsim/PhotonPropagation/PhotonVisibilityService.h
```diff
@@ -298,6 +300,27 @@ namespace phot {
...
 
+    void findVoxelSuggestion(std::string& logString,
+                             float tpcMin,
+                             float tpcMax,
+                             float cryoMin,
+                             float cryoMax,
+                             int& nVoxels,
+                             float& voxelMin,
+                             float& voxelMax,
+                             float voxelSizeGoal = 10.0) const;
```
knoepfel: Suggest removing the default value of 10.0 as it's not used anywhere--i.e. an explicit `voxelSizeGoal` argument is always specified when invoking `findVoxelSuggestion(...)`.

larsim/PhotonPropagation/PhotonVisibilityService.h
```diff
@@ -298,6 +300,27 @@ namespace phot {
...
+    float testVoxelSuggestion(std::string& logString,
+                              float tpcMin,
+                              float tpcMax,
+                              float cryoMin,
+                              float cryoMax,
+                              int& nVoxels,
+                              float& voxelMin,
+                              float& voxelMax,
+                              float voxelSizeGoal = 10.0,
+                              int jog = 0) const;
```
knoepfel: I suggest the following adjustment:

```suggestion
    float testVoxelSuggestion(float tpcMin,
                              float tpcMax,
                              float cryoMin,
                              float cryoMax,
                              int& nVoxels,
                              float& voxelMin,
                              float& voxelMax,
                              float voxelSizeGoal,
                              int jog,
                              std::string* logString = nullptr) const;
```

The `voxelSizeGoal` and `job` arguments are always specified when `testVoxelSuggestion(...)` is invoked.  And by making `logString` a pointer and giving it a default value of `nullptr`, you can avoid passing in dummy strings when there's nothing to log.

larsim/PhotonPropagation/PhotonVisibilityService.cc
```diff
@@ -831,4 +916,88 @@ namespace phot {
...
+                                                     float& voxelMax,
+                                                     float voxelSizeGoal,
+                                                     int jog) const
+  {
+    // Examine TPC to establish voxel size
+    float tpcSize = tpcMax - tpcMin;
+    int tpcVoxelNumber = ceil(tpcSize / voxelSizeGoal) + jog; // units are cm
+    float voxelSize = tpcSize / tpcVoxelNumber;
+    float cryoSize = cryoMax - cryoMin;
+    logString += "\tTPC size " + std::to_string(tpcSize) + "cm requires " +
```
knoepfel: Assuming you change the `testVoxelSuggestion(...)` function signature as suggested, you can then do:

```c++
if (logString) {
  *logString += ...
}

**Review by knoepfel — COMMENTED** [review]
@bwynneHEP, I've added some implementation suggestions for your consideration.  I'm happy to discuss any of them with you.

File: larsim/PhotonPropagation/PhotonVisibilityService.cc (hunk 5 above)
Comment (bwynneHEP) [reply larsim/PhotonPropagation/PhotonVisibilityService.cc]: Agreed, done

File: larsim/PhotonPropagation/PhotonVisibilityService.h (hunk 4 above)
Comment (bwynneHEP) [reply larsim/PhotonPropagation/PhotonVisibilityService.h]: Legacy of development, good call. Fixed.

File: larsim/PhotonPropagation/PhotonVisibilityService.h (hunk 3 above)
Comment (bwynneHEP) [reply larsim/PhotonPropagation/PhotonVisibilityService.h]: Fixed.

File: larsim/PhotonPropagation/PhotonVisibilityService.cc (hunk 1 above)
Comment (bwynneHEP) [reply larsim/PhotonPropagation/PhotonVisibilityService.cc]: Oops, been spending too much time with Python. Fixed.

File: larsim/PhotonPropagation/PhotonVisibilityService.cc (hunk 2 above)
Comment (bwynneHEP) [reply larsim/PhotonPropagation/PhotonVisibilityService.cc]: yup

**Comment by bwynneHEP** [comment]
Apologies for spam, minor git fail. Comments should now be incorporated, thanks @knoepfel

**Comment by knoepfel** [comment]
trigger build

**Comment by bwynneHEP** [comment]
> -uboone tests failed on slf7 for e26:prof for details see https://lar-ci-history.fnal.gov/LarCI/app/ns:uboone/view_builds/index?offset=0&builds=uboone_ci/10541&builds= for details of the parent CI build see https://lar-ci-history.fnal.gov/LarCI/app/view_builds/index?offset=0&builds=lar_ci/22554&builds=

The uBOONE fails seem to be unrelated to code changes

```
135: ----------- check this block for errors -----------------------
136: Error encountered when setting up product: larpandora
137: ERROR: Version conflict -- dependency tree requires versions conflicting with current setup of product larpandora: version v09_22_01 vs v09_22_04
138: ERROR: setup -B larpandora v09_22_01 -q e26:+prof failed
139: ERROR: setup of required products has failed
```

tomjunk:
Approved -- looks okay and defaults to no change in workflows.

lgarren:
approve

PR #136, opened 2024-03-27
Adding option to MUSUN generator to rotate coordinate system

vpec0:
This adds configurables  to MUSUN generator so that users can rotate the coordinate system to match the detector coordinate system.

By default, MUSUN assumes the standard DUNE Far Detector Horizontal Drift coordinate system, with Z in beam direction, Y vertical, and X in the horizontal drift direction. In order to be able to generate muons within Vertical Drift geometry which assumes X to be vertical, MUSUN's internal coordinate system needs to be rotated accordingly.

Example of how to configure this rotation for DUNE FD VD is included in comments within MUSUN.fcl.

This modification was tested long time ago, within DUNE Calibration Working Group, where cosmic muon sample was generated for the VD geometry.

### DIFF
--- a/larsim/EventGenerator/MuonPropagation/MUSUN.fcl
+++ b/larsim/EventGenerator/MuonPropagation/MUSUN.fcl
@@ -9,9 +9,9 @@ standard_MUSUN:
  PDG:                    -13        # PdG codes of particles to make
  ChargeRatio:            1.38       # Ratio of particle/anti-particle
                                     # For one particle type set equal to 0.
-                                    
- # Directory where ROOT pdf file is. If not using one of the standard configurations 
- # you will !!NEED!! to change this to your own directory!!!  
+
+ # Directory where ROOT pdf file is. If not using one of the standard configurations
+ # you will !!NEED!! to change this to your own directory!!!
  InputDir:               "/dune/data/users/warburton/MUSUN/"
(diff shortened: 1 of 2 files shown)


knoepfel:
@vpec0, just confirming that the default configuration does not change current behavior.  Also, why does the VD require a rotation around the Z axis?

vpec0:
Yes, I can confirm that the default configuration does not change the behaviour. The HD and VD use rotated coordinate systems, VD has X coordinate axis pointing upwards whereas HD has it point horizontally. @knoepfel

Viktor

knoepfel:
trigger build

lgarren:
@LArSoft/level-2-managers  please approve or comment

tomjunk (approved):
I'll go ahead and approve this so it can go into a release, but there is a concern about files in

/dune/data/users/warburton/MUSUN/

This directory is invisible on the grid, and will be invisible to everyone in May 2024.  If those files are useful, I recommend putting them in StashCache.  I will e-mail Viktor and Karl about this.

tomjunk:
Speaking of filenames, if you have an input filename that is supposed to be replaced and cannot possibly work, then choose a more obvious dummy name, like "/replace/this/with/a/valid/file" or something.  To point to a file that looks like it contains useful data but which no longer exists, may cause some users to worry.

lgarren:
approve

PR #142, opened 2024-10-03
Fixing GENIE GReWeightResonanceDecay Double Application

leehagaman:
@nitish-nayak and I found an issue that causes the GReWeightResonanceDecay calculator to be applied twice, leading to squared weights for the BR1gamma, BR1eta, and Theta_Delta2Npi knobs. This was recently fixed in GENIE, described [here](https://github.com/GENIE-MC/Reweight/pull/34).


knoepfel:
trigger build

tomjunk:
Thanks for spotting this! Steven Gardiner should know more about the knock-on effects of changing things for people who are currently running the buggy version and may get a mixture of buggy and non-buggy weights. Though they should use frozen releases if they need bug-for-bug compatibility, so I am almost always in favor of just fixing bugs and moving forwards.  Since Steven approved the one in GENIE, I'll take his word for it and approve this one.

tomjunk:
Approved

lgarren:
approve

PR #143, opened 2024-12-09
Add variables to set region of photon emission

szhang17phys:
User can invoke the new variables directly at the fhicl file. Then region of photon emission can be determined based on the interest of the user, which is more flexible than the initial module, where photon emission vertices can only be spread in the whole space.


lgarren:
code-checks

lgarren:
@szhang17phys please run clang format as noted

larsim/EventGenerator/PhotonGen_module.cc
```diff
@@ -110,6 +118,13 @@ namespace evgen {
     , fP{pset.get<double>("P")}
     , fSigmaP{pset.get<double>("SigmaP")}
     , fN{pset.get<int>("N")}
+    , fBminX{pset.get<double>("BminX")}
+    , fBmaxX{pset.get<double>("BmaxX")}
+    , fBminY{pset.get<double>("BminY")}
+    , fBmaxY{pset.get<double>("BmaxY")}
+    , fBminZ{pset.get<double>("BminZ")}
+    , fBmaxZ{pset.get<double>("BmaxZ")}
+
```
knoepfel: Two things:

1. I'd rename the configuration parameters (e.g.) `Xmin` instead of `BminX`.
2. Also, this change would require all users to specify `"Bmin*"` and `"Bmax*"` parameters in their configurations, and that is not always desired.  It would be better to make it an option--i.e. if the parameter `Ymin` is provided, then it should override the value of `fYmin`.

larsim/EventGenerator/PhotonGen_module.cc
```diff
@@ -152,9 +169,23 @@ namespace evgen {
...
     std::cout << "Xmin: " << fXmin << " Xmax: " << fXmax << " Ymin: " << fYmin << " Ymax: " << fYmax
               << " Zmin: " << fZmin << " Zmax: " << fZmax << std::endl;
+
+    fXmin = fBminX;
+    fXmax = fBmaxX;
+    fYmin = fBminY;
+    fYmax = fBmaxY;
+    fZmin = fBminZ;
+    fZmax = fBmaxZ;
```
knoepfel: See above comment.  These assignments should only be made when the configuration parameter is provided.  Otherwise, the values of `fXmin`, `fXmax`, etc. should be set to `CryoBounds.MinX()`, etc.

knoepfel (commented):
@szhang17phys, thank you for the PR.  Please see the comments below.

szhang17phys:
Thanks for above nice comments! I will modify accordingly!

szhang17phys (commented):
Names of new variables modified.

lgarren:
@szhang17phys please run the code formatting https://github.com/LArSoft/larsim/pull/143#issuecomment-2582480425

knoepfel (commented):
Thanks for the updates @szhang17phys --we're getting closer.  I was thinking of something like this:

```c++
PhotonGen::PhotonGen(fhicl::ParameterSet const& pset)
  : ...
{
  if (auto boundaries = pset.get_if_present<fhicl::ParameterSet>("boundaries")) {
    fXmin = boundaries->get<double>("Xmin");
    fXmax = boundaries->get<double>("Xmax");
    fYmin = boundaries->get<double>("Ymin");
    fYmax = boundaries->get<double>("Ymax");
    fZmin = boundaries->get<double>("Zmin");
    fZmax = boundaries->get<double>("Zmax");
  }
  else {
    art::ServiceHandle<geo::Geometry const> geo;
    auto const CryoBounds = geo->Cryostat().Boundaries();
    fXmin = CryoBounds.MinX();
    fXmax = CryoBounds.MaxX();
    fYmin = CryoBounds.MinY();
    fYmax = CryoBounds.MaxY();
    fZmin = CryoBounds.MinZ();
    fZmax = CryoBounds.MaxZ();
  }
}
```

where `boundaries` would be a nested table in the configuration:

```jsonnet
photonGen: {
  module_type: PhotonGen
  boundaries: {
    Xmin: ...
  }
}
```

If the `boundaries` parameter is in the configuration, then the module will take the parameters from the configuration; otherwise they will be provided by the geometry system.  This involves the following changes:

1. Removing the `fUserD`, `fBmin*`, and `fBmax*` variables, and
2. Moving the setting of the `fXmin` (etc.) variables to the constructor.

Let me know if you have any questions.

szhang17phys:
Hello! Kyle. I think I got your idea: You do not want to introduce too many variables. fXmin* will replace my fBminX* in PhotonGen::PhotonGen(fhicl::ParameterSet const& pset). If fXmin, fXmax, ... are not pointed to certain values by user, they will get values from cryoBounds automatically. I agree that your proposal is more neat! I will modify again recently!

PR #144, opened 2024-12-11
Update SingleGen_module.cc to fix seed generation bug.

mantheys:
The seed was always set to 0. So 2 different calls to the SinglesGen_module where generating the same random output.

### DIFF
--- a/larsim/EventGenerator/SingleGen_module.cc
+++ b/larsim/EventGenerator/SingleGen_module.cc
@@ -177,7 +177,7 @@ namespace evgen {
       fhicl::Atom<rndm::NuRandomService::seed_t> Seed{
         Name("Seed"),
         Comment("override the random number generator seed"),
-        0 /* dummy? */};
+        rndm::NuRandomService::InvalidSeed};
 
     private:
       /// Returns whether the specified mode is an histogram distribution.
--- a/larsim/EventGenerator/SingleGen_module.cc
+++ b/larsim/EventGenerator/SingleGen_module.cc
@@ -173,7 +173,7 @@ namespace evgen {
         Name("ThetaXzYzHist"),
         Comment("name of the histograms of angular (X-Z and Y-Z) distribution"),
         [this]() { return fromHistogram(AngleDist()); }};
-      
+
       fhicl::Atom<rndm::NuRandomService::seed_t> Seed{
         Name("Seed"),
         Comment("override the random number generator seed"),
--- a/larsim/EventGenerator/SingleGen_module.cc
+++ b/larsim/EventGenerator/SingleGen_module.cc
@@ -447,7 +447,9 @@ namespace evgen {
     , fHistFileName(config().HistogramFile())
     , fPHist(config().PHist())
     , fThetaXzYzHist(config().ThetaXzYzHist())
-    , fEngine(art::ServiceHandle<rndm::NuRandomService>()->registerAndSeedEngine(createEngine(0), "HepJamesRandom", "Seed"))
+    , fEngine(art::ServiceHandle<rndm::NuRandomService>()->registerAndSeedEngine(createEngine(0),
+                                                                                 "HepJamesRandom",
+                                                                                 "Seed"))
   {
     setup();
     rndm::NuRandomService::seed_t seed;
--- a/larsim/EventGenerator/SingleGen_module.cc
+++ b/larsim/EventGenerator/SingleGen_module.cc
@@ -173,7 +173,7 @@ namespace evgen {
         Name("ThetaXzYzHist"),
         Comment("name of the histograms of angular (X-Z and Y-Z) distribution"),
         [this]() { return fromHistogram(AngleDist()); }};
-
+      
       fhicl::Atom<rndm::NuRandomService::seed_t> Seed{
         Name("Seed"),
         Comment("override the random number generator seed"),
@@ -448,9 +448,8 @@ namespace evgen {
     , fHistFileName(config().HistogramFile())
     , fPHist(config().PHist())
     , fThetaXzYzHist(config().ThetaXzYzHist())
-    , fEngine(art::ServiceHandle<rndm::NuRandomService>()->registerAndSeedEngine(createEngine(config().Seed()),
-                                                                                 "HepJamesRandom",
-                                                                                 "Seed"))
+    , fEngine(art::ServiceHandle<rndm::NuRandomService>()
+                ->registerAndSeedEngine(createEngine(config().Seed()), "HepJamesRandom", "Seed"))
   {
     setup();
     mf::LogInfo("SingleGen") << "Seed set to " << fEngine.getSeed();
--- a/larsim/EventGenerator/SingleGen_module.cc
+++ b/larsim/EventGenerator/SingleGen_module.cc
@@ -174,9 +174,10 @@ namespace evgen {
         Comment("name of the histograms of angular (X-Z and Y-Z) distribution"),
         [this]() { return fromHistogram(AngleDist()); }};
 
-      fhicl::OptionalAtom<rndm::NuRandomService::seed_t> Seed{
+      fhicl::Atom<rndm::NuRandomService::seed_t> Seed{
         Name("Seed"),
-        Comment("override the random number generator seed")};
+        Comment("override the random number generator seed"),
+        0 /* dummy? */};
 
     private:
       /// Returns whether the specified mode is an histogram distribution.
@@ -447,20 +448,12 @@ namespace evgen {
     , fHistFileName(config().HistogramFile())
     , fPHist(config().PHist())
     , fThetaXzYzHist(config().ThetaXzYzHist())
-    , fEngine(art::ServiceHandle<rndm::NuRandomService>()->registerAndSeedEngine(createEngine(0),
+    , fEngine(art::ServiceHandle<rndm::NuRandomService>()->registerAndSeedEngine(createEngine(config().Seed()),
                                                                                  "HepJamesRandom",
                                                                                  "Seed"))
   {
     setup();
-    rndm::NuRandomService::seed_t seed;
-    if (config().Seed(seed)) {
-      fEngine.setSeed(seed, 0 /* dummy? */);
-      mf::LogInfo("SingleGen") << "Seed set to " << seed << " (from configuration)";
-    }
-    else {
-      seed = fEngine.getSeed();
-      mf::LogInfo("SingleGen") << "Seed set to " << seed;
-    }
+    mf::LogInfo("SingleGen") << "Seed set to " << fEngine.getSeed();
 
     produces<std::vector<simb::MCTruth>>();
     produces<sumdata::RunData, art::InRun>();
--- a/larsim/EventGenerator/SingleGen_module.cc
+++ b/larsim/EventGenerator/SingleGen_module.cc
@@ -449,7 +449,7 @@ namespace evgen {
     , fPHist(config().PHist())
     , fThetaXzYzHist(config().ThetaXzYzHist())
     , fEngine(art::ServiceHandle<rndm::NuRandomService>()
-                ->registerAndSeedEngine(createEngine(config().Seed()), "HepJamesRandom", "Seed"))
+                ->registerAndSeedEngine(createEngine(config().Seed()), "", "", config().Seed()))
   {
     setup();
     mf::LogInfo("SingleGen") << "Seed set to " << fEngine.getSeed();
--- a/larsim/EventGenerator/SingleGen_module.cc
+++ b/larsim/EventGenerator/SingleGen_module.cc
@@ -447,11 +447,18 @@ namespace evgen {
     , fHistFileName(config().HistogramFile())
     , fPHist(config().PHist())
     , fThetaXzYzHist(config().ThetaXzYzHist())
-    , fEngine(createEngine(0))
+    , fEngine(art::ServiceHandle<rndm::NuRandomService>()->registerAndSeedEngine(createEngine(0), "HepJamesRandom", "Seed"))
   {
     setup();
     rndm::NuRandomService::seed_t seed;
-    if (config().Seed(seed)) { fEngine.setSeed(seed, 0 /* dummy? */); }
+    if (config().Seed(seed)) {
+      fEngine.setSeed(seed, 0 /* dummy? */);
+      mf::LogInfo("SingleGen") << "Seed set to " << seed << " (from configuration)";
+    }
+    else {
+      seed = fEngine.getSeed();
+      mf::LogInfo("SingleGen") << "Seed set to " << seed;
+    }
 
     produces<std::vector<simb::MCTruth>>();
     produces<sumdata::RunData, art::InRun>();


mantheys:
@absolution1

lgarren:
@mantheys please run clang format as suggested above

larsim/EventGenerator/SingleGen_module.cc
```diff
@@ -447,11 +447,20 @@ namespace evgen {
...
     rndm::NuRandomService::seed_t seed;
-    if (config().Seed(seed)) { fEngine.setSeed(seed, 0 /* dummy? */); }
+    if (config().Seed(seed)) {
+      fEngine.setSeed(seed, 0 /* dummy? */);
+      mf::LogInfo("SingleGen") << "Seed set to " << seed << " (from configuration)";
+    }
+    else {
+      seed = fEngine.getSeed();
+      mf::LogInfo("SingleGen") << "Seed set to " << seed;
+    }
```
absolution1: @mantheys 
Can you confirm what printout you now see in, for example, a DUNE far detector particle gun job?

absolution1 (commented):
Thanks @mantheys 
I had one quick question about printout you now see after the fix.

larsim/EventGenerator/SingleGen_module.cc
mantheys: Sure! This is the full output for a single electron production in DUNE larsoft (https://github.com/DUNE/dunesw/blob/develop/fcl/dunefd/gen/solar/prodelectron_solar_flat_dune10kt_1x2x6.fcl). The relevant part of the output start at line 65.
```
%MSG-i RANDOM:  SingleGen:generator@Construction 12-Dec-2024 12:07:01 CET  ModuleConstruction
Instantiated HepJamesRandom engine "generator:0:" with seed 0.
%MSG
%MSG-i NuRandomService:  SingleGen:generator@Construction  12-Dec-2024 12:07:01 CET ModuleConstruction
Seeding HepJamesRandom engine "generator:Seed" with seed 575704593.
%MSG
%MSG-i SingleGen:  SingleGen:generator@Construction 12-Dec-2024 12:07:01 CET  ModuleConstruction
Seed set to 575704593
```
[SinglesGen.txt](https://github.com/user-attachments/files/18110485/SinglesGen.txt)

absolution1: Thanks for this Sergio.  So it looks like the code is now initialising a sensible seed when not using the config.
Can you confirm that you get different seeds between two jobs?

mantheys: Yes exactly! That's the behavior I see now.

absolution1: Perfect, thank you!

absolution1:
Hi @knoepfel @lgarren 
This PR looks completely sensible to me, and I am happy to approve.
Is the engine setup in keeping with LArSoft standards?

lgarren:
@mantheys Is this a fix for https://cdcvs.fnal.gov/redmine/issues/28411?

mantheys:
Hi @lgarren, I'm not in contact with the people who opened the issue you are referring to; but from the description, it looks to me exactly like the problem I was facing. The Seed was always set to 0, independently of the NuRandomService "policy" one decided to configure in the fcl. My PR at least fixed the case in which one wants the policy to be "random" so that each time one runs the simulation the Seed is set to a different value.

knoepfel (commented):
These changes look good, @mantheys.  I do have one suggestion that should make the code a bit cleaner.  First, I'd change the configuration parameter `Seed` to be an `Atom` with a default value and not an `OptionalAtom`:

```diff
-       fhicl::OptionalAtom<rndm::NuRandomService::seed_t> Seed{
+       fhicl::Atom<rndm::NuRandomService::seed_t> Seed{
          Name("Seed"),
          Comment("override the random number generator seed")
+         , 0 /* dummy? */
        };
```

You can then change the rest of the code to be rather simple:

```C++
  ...
  , fEngine{
      art::ServiceHandle<rndm::NuRandomService>()->registerAndSeedEngine(
        createEngine(config().Seed()),
        "HepJamesRandom")}
  {
    setup();
    mf::LogInfo("SingleGen") << "Seed set to " << fEngine.getSeed();
  }
```

larsim/EventGenerator/SingleGen_module.cc
```diff
@@ -447,11 +447,20 @@ namespace evgen {
     , fHistFileName(config().HistogramFile())
     , fPHist(config().PHist())
     , fThetaXzYzHist(config().ThetaXzYzHist())
-    , fEngine(createEngine(0))
+    , fEngine(art::ServiceHandle<rndm::NuRandomService>()->registerAndSeedEngine(createEngine(0),
+                                                                                 "HepJamesRandom",
+                                                                                 "Seed"))
```
PetrilloAtWork: Why not to use [`registerAndSeedEngine(engine_t& engine, std::string type, SeedAtom const&`](https://code-doc.larsoft.org/docs/latest/html/classrndm_1_1NuRandomService.html#a70a3880022efe4b02b356b0dd84ce0d5):
```suggestion
    , fEngine(art::ServiceHandle<rndm::NuRandomService>()->registerAndSeedEngine(createEngine(0),
                                                                                 "HepJamesRandom",
                                                                                 config().Seed))
```
or [the variant with `std::optional` seed](https://code-doc.larsoft.org/docs/latest/html/classrndm_1_1NuRandomService.html#aed73c4bd19a744678a37579e5b433b65):
```suggestion
    , fEngine(art::ServiceHandle<rndm::NuRandomService>()->registerAndSeedEngine(createEngine(0),
                                                                                 "HepJamesRandom",
                                                                                 "",
                                                                                 config().Seed()))
```
In this way the initialisation of the seed should happen at once.

larsim/EventGenerator/SingleGen_module.cc
PetrilloAtWork: There may be a problem here.
The engine is directly seeded, bypassing `NuRandomService`.
`NuRandomService` does not know that the seed was set manually and should therefore be considered frozen, and it will happily override it if the configured policy does that.

On the other end, using `registerAndSeedEngine()` will implicitly freeze the seed if one is found in the optional parameter specified in the argument.

`NuRandomService` can also print the seeds as used — there are [configuration parameters](https://code-doc.larsoft.org/docs/latest/html/classrndm_1_1SeedMaster.html#details) to make it do that: `services.NuRandomService.endOfJobSummary: true` and/or a non-zero value for `services.NuRandomService.verbose`.
This service should be rewritten in modern _art_ (configuration validation and _art_ tools)... heh.

PetrilloAtWork (commented):
I have a concern with the initialisation approach, that I described in the second comment.

larsim/EventGenerator/SingleGen_module.cc
mantheys: Hi @PetrilloAtWork, sorry if I'm not following the conversation correctly. Do you agree with @knoepfel's proposed solution or are you suggesting an alternative approach? Thanks in advance!

absolution1 (commented):
I'm a little bit confused by the change.  Does this now mean that the seed has to be read from fcl?

knoepfel:
@absolution1, this means that a default FHiCL value of 0 is now provided for the seed.  This has identical semantics to what was implemented earlier, it's just that it's done more concisely.

absolution1:
Does the change allow for letting the NuRandomService generate the seed?

larsim/EventGenerator/SingleGen_module.cc
PetrilloAtWork: We happened to be writing something similar at the same time, unaware of each other.

However I have to check whether there is a difference between telling `NuRandomService` to use seed `0` and not specifying any seed. Chances are that the former would force the seed to be `0`, disabling `NuRandomService` seed management.

PetrilloAtWork (changes requested):
This change will set the `Seed` parameter to `0` if it is not specified in the configuration. `NuRandomService` protocol prescribes that if the specified seed value is not `rndm::NuRandomService::InvalidSeed`, that value will be used to _fix_ the seed (see "Overriding the seed from `rndm::NuRandomService` at run time" section of [`rndm::NuRandomService` details documentation](https://code-doc.larsoft.org/docs/latest/html/classrndm_1_1NuRandomService.html#details), unfortunately not well Doxygen—written). In case the seed is not specified, then, if `0` is different from `rndm::NuRandomService::InvalidSeed`, the seed will be fixed to the value `0`.

Now it happens that `rndm::NuRandomService::InvalidSeed` is actually implemented to be `0`, but we should not rely on that implementation detail.
My recommendation is to use a configuration parameter of type [`rndm::SeedAtom`](https://code-doc.larsoft.org/docs/latest/html/namespacerndm.html#a8ca1c9d6cc14c6f4cb1cc0384f7899d4) (which is under the hood a `fhicl::OptionalAtom`) and let `NuRandomEngine::registerAndSeedEngine()` handle the presence or absence of the seed; this approach moves almost all the maintenance out of the module (this shields it behind a service protocol). If this is not desired, then I recommend to replace the default value using `rndm::NuRandomService::InvalidSeed` instead of its value `0`.

PetrilloAtWork:
> Does the change allow for letting the NuRandomService generate the seed?

I had to do some code investigation; the answer is that yes, it does because the default value chosen here, `0`, "happens to be" a special value that tells the service to take care of the seed.
I am nevertheless requesting a change that makes this more evident to the reader of the module code.

larsim/EventGenerator/SingleGen_module.cc
```diff
@@ -447,11 +448,11 @@ namespace evgen {
     , fHistFileName(config().HistogramFile())
     , fPHist(config().PHist())
     , fThetaXzYzHist(config().ThetaXzYzHist())
-    , fEngine(createEngine(0))
+    , fEngine(art::ServiceHandle<rndm::NuRandomService>()
+                ->registerAndSeedEngine(createEngine(config().Seed()), "HepJamesRandom", "Seed"))
```
PetrilloAtWork:
```suggestion
    , fEngine(art::ServiceHandle<rndm::NuRandomService>()
                ->registerAndSeedEngine(createEngine(config().Seed()), "HepJamesRandom", "Seed"))
```
Note that `config().Seed()` in the `createEngine()` argument is always going to be overwritten, either by the same value again, or by the seed assigned to the engine by `NuRandomService`.
Whether having `config().Seed()` is more or less confusing than having `0` as argument is up to the reader (I prefer the latter because it's "clear" that it can't be the final setting, but again that's just me).

larsim/EventGenerator/SingleGen_module.cc
```diff
@@ -173,10 +173,11 @@ namespace evgen {
...
         Comment("name of the histograms of angular (X-Z and Y-Z) distribution"),
         [this]() { return fromHistogram(AngleDist()); }};
-
-      fhicl::OptionalAtom<rndm::NuRandomService::seed_t> Seed{
+      
+      fhicl::Atom<rndm::NuRandomService::seed_t> Seed{
         Name("Seed"),
-        Comment("override the random number generator seed")};
+        Comment("override the random number generator seed"),
+        0 /* dummy? */};
```
knoepfel: @mantheys, based on @PetrilloAtWork's comment, I suggest:

```suggestion
        rndm::NuRandomService::InvalidSeed};
```

With that change, I would consider this PR ready.

larsim/EventGenerator/SingleGen_module.cc
knoepfel: Thanks for looking into this, @PetrilloAtWork.  I suggested that @mantheys use `InvalidSeed` as the default parameter value.

larsim/EventGenerator/SingleGen_module.cc
mantheys: Done!

knoepfel:
code-checks

knoepfel:
trigger build

lgarren:
@mantheys @knoepfel it looks as though at least one of the DUNE CI failures might be related to this PR.

knoepfel:
@lgarren, the error looks like a WireCell problem:

```
%MSG-s ArtException:  WireCellToolkit:wclsdatasp@Construction  19-Dec-2024 09:59:21 CST ModuleConstruction
60: cet::exception caught in art
61: ---- Configuration BEGIN
62:   make_tool: 
63:   ---- WireCellLArSoft BEGIN
64:     ../../source/wirecell-0.29.4/gen/src/AnodePlane.cxx(167): Throw in function configure
65:     Dynamic exception type: boost::wrapexcept<WireCell::ValueError>
66:     [tag_stacktrace*] =  0# WireCell::Gen::AnodePlane::configure(Json::Value const&) in /cvmfs/larsoft.opensciencegrid.org/products/wirecell/v0_29_4/Linux64bit+3.10-2.17-e26-prof/lib/libWireCellGen.so
67:      1# WireCell::Main::initialize() in /cvmfs/larsoft.opensciencegrid.org/products/wirecell/v0_29_4/Linux64bit+3.10-2.17-e26-prof/lib/libWireCellApps.so
68:      2# wcls::WCLS::WCLS(art::ToolConfigTable<wcls::WCLSConfig, wcls::WCLSKeysToIgnore> const&) in /cvmfs/larsoft-ib.opensciencegrid.org/LArSoft/LArSoft_custom_slf7_e26_prof_20241219_092815/localProducts_LArSoft_LArSoft_lar_ci_e26_prof/larwirecell/v10_00_05/slf7.x86_64.e26.prof/lib/liblarwirecell_Tools_WCLS_tool.so
69:      3# std::__detail::_MakeUniq<wcls::WCLS>::__single_object std::make_unique<wcls::WCLS, fhicl::ParameterSet const&>(fhicl::ParameterSet const&) in /cvmfs/larsoft-ib.opensciencegrid.org/LArSoft/LArSoft_custom_slf7_e26_prof_20241219_092815/localProducts_LArSoft_LArSoft_lar_ci_e26_prof/larwirecell/v10_00_05/slf7.x86_64.e26.prof/lib/liblarwirecell_Tools_WCLS_tool.so
70:      4# makeTool in /cvmfs/larsoft-ib.opensciencegrid.org/LArSoft/LArSoft_custom_slf7_e26_prof_20241219_092815/localProducts_LArSoft_LArSoft_lar_ci_e26_prof/larwirecell/v10_00_05/slf7.x86_64.e26.prof/lib/liblarwirecell_Tools_WCLS_tool.so
71:      5# std::enable_if<std::is_class<wcls::MainTool>::value, art::detail::tool_type<wcls::MainTool, void>::return_type>::type art::make_tool<wcls::MainTool>(fhicl::ParameterSet const&) in /cvmfs/larsoft-ib.opensciencegrid.org/LArSoft/LArSoft_custom_slf7_e26_prof_20241219_092815/localProducts_LArSoft_LArSoft_lar_ci_e26_prof/larwirecell/v10_00_05/slf7.x86_64.e26.prof/lib/liblarwirecell_Modules_WireCellToolkit_module.so
72:      6# wcls::WireCellToolkit::reconfigure(fhicl::ParameterSet const&) in /cvmfs/larsoft-ib.opensciencegrid.org/LArSoft/LArSoft_custom_slf7_e26_prof_20241219_092815/localProducts_LArSoft_LArSoft_lar_ci_e26_prof/larwirecell/v10_00_05/slf7.x86_64.e26.prof/lib/liblarwirecell_Modules_WireCellToolkit_module.so
73:      7# wcls::WireCellToolkit::WireCellToolkit(fhicl::ParameterSet const&, art::ProcessingFrame const&) in /cvmfs/larsoft-ib.opensciencegrid.org/LArSoft/LArSoft_custom_slf7_e26_prof_20241219_092815/localProducts_LArSoft_LArSoft_lar_ci_e26_prof/larwirecell/v10_00_05/slf7.x86_64.e26.prof/lib/liblarwirecell_Modules_WireCellToolkit_module.so
74:      8# make_module in /cvmfs/larsoft-ib.opensciencegrid.org/LArSoft/LArSoft_custom_slf7_e26_prof_20241219_092815/localProducts_LArSoft_LArSoft_lar_ci_e26_prof/larwirecell/v10_00_05/slf7.x86_64.e26.prof/lib/liblarwirecell_Modules_WireCellToolkit_module.so
75:      9# art::PathManager::makeModule_[abi:cxx11](fhicl::ParameterSet const&, art::ModuleDescription const&, art::ScheduleID) const in /cvmfs/larsoft.opensciencegrid.org/products/art/v3_14_04/slf7.x86_64.e26.prof/lib/libart_Framework_Core.so
76:     10# art::PathManager::makeModules_(unsigned short) in /cvmfs/larsoft.opensciencegrid.org/products/art/v3_14_04/slf7.x86_64.e26.prof/lib/libart_Framework_Core.so
77:     11# art::PathManager::createModulesAndWorkers(art::GlobalTaskGroup&, art::detail::SharedResources&, std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > const&) in /cvmfs/larsoft.opensciencegrid.org/products/art/v3_14_04/slf7.x86_64.e26.prof/lib/libart_Framework_Core.so
78:     12# art::EventProcessor::EventProcessor(fhicl::ParameterSet, art::detail::EnabledModules) in /cvmfs/larsoft.opensciencegrid.org/products/art/v3_14_04/slf7.x86_64.e26.prof/lib/libart_Framework_EventProcessor.so
79:     13# art::run_art(int, char**, boost::program_options::options_description&, std::vector<std::unique_ptr<art::OptionsHandler, std::default_delete<art::OptionsHandler> >, std::allocator<std::unique_ptr<art::OptionsHandler, std::default_delete<art::OptionsHandler> > > >&&) in /cvmfs/larsoft.opensciencegrid.org/products/art/v3_14_04/slf7.x86_64.e26.prof/lib/libart_Framework_Art.so
80:     14# artapp(int, char**, bool) in /cvmfs/larsoft.opensciencegrid.org/products/art/v3_14_04/slf7.x86_64.e26.prof/lib/libart_Framework_Art.so
81:     15# main in lar
82:     16# __libc_start_main in /lib64/libc.so.6
83:     17# 0x00000000004012D0 in lar
84:     
85:     [WireCell::tag_errmsg*] = AnodePlane: error in configuration, expect scalar values for response_x and anode_x.
86:   ---- WireCellLArSoft END
87:   Exception caught while processing plugin spec: WCLS
88: ---- Configuration END
89: %MSG
```

mantheys:
This error typically occurs when someone does not run "mrbslp" before executing a lar -c command. Not sure how this could have anything to do with the bug we are solving.

lgarren:
@mantheys @knoepfel There is definitely a wirecell problem, but @tomjunk thinks there is also a problem related to this PR.
See ci_gen_regression_test_protoDUNEdp

```
134: %MSG
135: %MSG-i SignalShapingServiceDUNE:  Early 19-Dec-2024 09:58:57 CST JobSetup
136: Getting Field response from .fcl file
137: %MSG
138: %MSG-i SignalShapingServiceDUNE:  Early 19-Dec-2024 09:58:57 CST JobSetup
139: Getting Filter from .fcl file
140: %MSG
141: %MSG-i RANDOM:  SingleGen:generator@Construction 19-Dec-2024 09:58:57 CST  ModuleConstruction
142: Instantiated HepJamesRandom engine "generator:0:" with seed 0.
143: %MSG
144: %MSG-s ArtException:  SingleGen:generator@Construction 19-Dec-2024 09:58:57 CST  ModuleConstruction
145: cet::exception caught in art
146: ---- Configuration BEGIN
147:   A seed for 'generator.Seed' was requested, but the configuration sets a nameless instance of 'generator'.
148:   Nameless and named engine instances can't coexist.
149: ---- Configuration END
150: %MSG
151: Art has completed and will exit with status 9.
152: 
```

larsim/EventGenerator/SingleGen_module.cc
```diff
@@ -447,11 +448,11 @@ namespace evgen {
     , fHistFileName(config().HistogramFile())
     , fPHist(config().PHist())
     , fThetaXzYzHist(config().ThetaXzYzHist())
-    , fEngine(createEngine(0))
+    , fEngine(art::ServiceHandle<rndm::NuRandomService>()
```
tomjunk, line 451: I normally copy an example from another module to get something working.  For example, is this pattern good enough?

https://github.com/LArSoft/larsim/blob/0e68c0382180e6d23049d3760d7ff8aa86841ba8/larsim/ElectronDrift/SimDriftElectrons_module.cc#L156

You can test your code interactively with a DUNE build, or just set up the DUNE stack with a build of larsim with your PR built, and run

ci_test_gen_protoDUNEdp.fcl

You may need to fetch a seed file to run this CI test fcl.  From the log:

https://dbweb8.fnal.gov:8443/LarCI/app/ns:dune/storage/docs/2024/12/19/stdout%23Sl9vs2k.log

7: fetch_files input /pnfs/dune/persistent/stash/ContinuousIntegration//protoDUNEdp/gen/GenRandomSeeds_Ref_protoDUNE.dat
8: 
9: Command: ifdh cp -D http://fndca1.fnal.gov:8000/pnfs/fnal.gov/usr/dune/persistent/stash/ContinuousIntegration//protoDUNEdp/gen/GenRandomSeeds_Ref_protoDUNE.dat ./

tomjunk:
Yes, two of our tests, the ProtoDUNE-DP and ProtoDUNE-VD gen tests failed with single gen seed configuration errors.  I put a comment on the line of code in the PR that is likely to be doing this.

knoepfel:
Re. https://github.com/LArSoft/larsim/pull/144#issuecomment-2555533867, yeah, we need to figure out what's going on there.  @mantheys, I can help with figuring it out in the new year—sorry I can't do it sooner.

mantheys:
Hi all! The original solution I proposed in the PR (https://github.com/LArSoft/larsim/pull/144/commits/365831a442a0d0d9ba36007730591405ea17afd3) was copied from another fcl and did in fact work on my local larsoft build producing the desired results. From what I can see it is very similar to what @tomjunk proposes in https://github.com/LArSoft/larsim/blob/0e68c0382180e6d23049d3760d7ff8aa86841ba8/larsim/ElectronDrift/SimDriftElectrons_module.cc#L156 If you all agree I would suggest to undo the last commits and only apply modifications that change the syntax and not the methods and perform again the CI test.

PetrilloAtWork:
There is a bit of confusion between two concepts.

1. The name of the instance of the random engine: random engines are specific to a module instance (e.g. `singles` instance of `SingleGen` module). To allow a module to use more than one random engine, e.g. to extract independently particle energy and position, _art_ assigns each random engine an "instance name"; so in this example that `singles` instance of module `SingleGen` can have a engine with instance name `"pos"` and one `"en"`, initialised with `createEngine(posSeed, "pos")` and `createEngine(enSeed, "en")`, and use them accordingly. Most of the times only one random engine is needed, and a single instance is used; the default instance name in this case is an empty one (`""`).
2. The name of the configuration key used to specify a seed for an engine; this name is determined when reading the configuration (`pset.get<seed_t>("Seed")` asks for a parameter with key `Seed`) or in the validated configuration object definition (`fhicl::Atom<seed_t>{ fhicl::Name{ "Seed" } }`). This name can't be empty. The name "Seed" here is not anything special, but it's a "tradition" for modules that use only one random engine instance.

The code in the current branch head, `registerAndSeedEngine(createEngine(config().Seed()), "HepJamesRandom", "Seed"))` performs two actions in sequence.

First, it creates a random engine with a default instance name (`createEngine(config().Seed())`) and sets an initial seed according to the configuration.

Then (`registerAndSeedEngine(..., "HepJamesRandom", "Seed")`) it registers that engine into `NuRandomService` as if it had instance name `"Seed"`, and it does not set its seed accordingly to the configuration at all (which is bad because then `NuRandomService` assumes that it can freely change the seed).

What is needed instead is for this second step to register the engine with the default empty name, and then set (again) the seed:
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~.cpp
registerAndSeedEngine(createEngine(config().Seed()), "", "", config().Seed()))
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
The first empty string is the type of engine, which is ignored; the second one is the random engine instance name, which matches the empty one implicitly used in the `createEngine()` call. The last parameter is the seed to be used; note that this is a `std::optional<seed_t>` argument and we are passing a `seed_t`: if the compiler can't cope, we'll need to convert it explicitly: `registerAndSeedEngine(createEngine(config().Seed()), "", "", std::optional{ config().Seed() }))` should do.
While setting the seed again in `registerAndSeedEngine()` after having set it with `createEngine()` has no immediate effect, it does have the essential side effect that `NuRandomService` is informed whether the seed is frozen by the configuration parameter or if it is free to manage (which happens if the seed value is `InvalidSeed`, as discussed elsewhere).

larsim/EventGenerator/SingleGen_module.cc
PetrilloAtWork: That example is not good enough, since it reads the initial seed from a free-form `fhicl::ParameterSet`, while `SingleGen` uses a validated configuration class.

larsim/EventGenerator/SingleGen_module.cc
knoepfel: @mantheys, let's try @PetrilloAtWork's suggestion:

```suggestion
                ->registerAndSeedEngine(createEngine(config().Seed()), "", "", config().Seed()))
```

knoepfel:
trigger build

knoepfel:
Approve

@mantheys, the CI tests look reasonable.  Thanks for your patience and willingness to sort out all of the above discussions.

PR #145, opened 2025-01-10
Updates to semi-analytical light simulation for DUNE 10kt simulation

pgreen135:
Adds two minor updates to semi-analytical light simulation to enable it to work in DUNE 10kt simulation (HD):

1. Adds temporary solution for identifying which photon detectors are visible depending on where the light emission occurs given the multiple optically isolated volumes in the DUNE 10kt geometry. Longer term, this is to be updated with a potential geometry service implementation currently under discussion.
2. Adds maximum distance between photon detectors and light emission to evaluate semi-analytical model, configured via fhicl parameter. Negligible light seen from very large distances, so calculations unnecessary -- leads to significant speed up in large geometries. 

This has been tested in DUNE validation system and sample of O(10k) test events successfully produced with 10kt geometry.

### DIFF
--- a/larsim/PhotonPropagation/PDFastSimPAR_module.cc
+++ b/larsim/PhotonPropagation/PDFastSimPAR_module.cc
@@ -302,8 +302,9 @@ namespace phot {
 
     // determine drift distance
     fDriftDistance = fGeom.TPC().DriftDistance();
-    // for multiple TPCs, use second TPC to skip small volume at edges of detector (DUNE)  
-    if (fNTPC > 1 && fDriftDistance < 50) fDriftDistance = fGeom.TPC(geo::TPCID{0, 1}).DriftDistance();
+    // for multiple TPCs, use second TPC to skip small volume at edges of detector (DUNE)
+    if (fNTPC > 1 && fDriftDistance < 50)
+      fDriftDistance = fGeom.TPC(geo::TPCID{0, 1}).DriftDistance();
 
     mf::LogInfo("PDFastSimPAR") << "PDFastSimPAR Initialization finish.\n"
                                 << "Simulate using semi-analytic model for number of hits."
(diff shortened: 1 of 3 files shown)


absolution1:
trigger build

absolution1:
The DUNE CI is reporting difference in the number of opdetwaveforms but I think that predates this PR

absolution1:
approve

pgreen135:
@absolution1 summarising the outcomes of the discussion we just had with @erica-snider and @knoepfel (correct me if I've misunderstood anything!):

- There isn't currently the possibility to have optical couplings between points and photodetectors in the geometry, and this is not something trivially added / something that necessarily even makes sense with the geometry structure. 
- We discussed some ideas on how this information could be accessed through creating some sort of mapping. However, given the use cases are fairly limited, this is probably overkill to try to have some from of detector agnostic method involving geometry/gdml. The more "hacky" approaches are probably sufficient, unless/until we find a case where these are no longer possible.
-  Instead, the suggestion was to elevate the hacky approach to a tool. This can then have a default that returns true (which would be sufficient for a most of the detectors), then for DUNE 10kt and SBND we have specific configurations of this that for now at least implement the existing hacky approaches. These would then be more easy to swap out / update / expand as needed for more geometries.

As for this PR, this then depends on timeline --- if this is needed immediately, this could be merged as-is. But the preference would be to create this tool instead. This isn't massively complicated, but requires my availability to do it and isn't something I'm likely to be able to do until after the CM/Workshop.

absolution1:
Hi @pgreen135 
Is there any downside to merging this as-is?  Are other experiments negatively affected?

pgreen135:
@absolution1 no, it shouldn't impact anything else -- the SBND implementation is unchanged

absolution1:
Thanks @pgreen135 
In that case, it would be good to go ahead with merging now, and add the tool as a later enhancement.

lgarren:
approve

PR #147, opened 2025-01-17
Ability to apply user-defined photon emission region added

szhang17phys:
I used the default variables fXmin, fXmax, etc to set the user-defined photon emission region. If these variables are not given new values by user, the module will set the values of these variables from the cryostat, i.e the default cryostat boundary. If user points values to these variables, user-defined variables will be taken.

### DIFF
--- a/larsim/EventGenerator/PhotonGen_module.cc
+++ b/larsim/EventGenerator/PhotonGen_module.cc
@@ -87,20 +87,8 @@ namespace evgen {
     double fP;      // central momentm of photon
     double fSigmaP; // mom width;
 
-    // Number of photons per event
+    //Number of photons per event
     int fN; // number of photons per event
-
-    bool fUserD; //Whether apply user-defined region: false = no, true = yes
-
-    //user-defined boundaries to constrain photon emission vertex---
-    double fBminX;
-    double fBmaxX;
-    double fBminY;
-    double fBmaxY;
-    double fBminZ;
-    double fBmaxZ;
-
-    CLHEP::HepRandomEngine& fEngine;
 
     //Boundaries of the detector
     double fXmin;
@@ -109,6 +97,8 @@ namespace evgen {
     double fYmax;
     double fZmin;
     double fZmax;
+
+    CLHEP::HepRandomEngine& fEngine;    
   };
 
   //----------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


knoepfel:
@szhang17phys, this is the right idea, but I think we need to be a little more careful.  Right now, it's only checking that the `Xmin` variable is something other than 0.  Suppose the user specifies only `Xmin: 1` in their configuration file...then every other parameter will default to 0, which we don't want.  Instead, see the proposal I made at https://github.com/LArSoft/larsim/pull/143#pullrequestreview-2551184113.

In that proposal, if the user wants to specify boundaries in the configuration, they have to specify the table `boundaries`, and then they have to specify *all* of the `Xmin`, `Xmax` (etc.) variables.

szhang17phys:
Hello! Kyle. It sounds like a good idea. I wonder how can I define and initialize the nested table boundaries? Is it correct to define the "boundaries" inside the private namespace? Moreover, Shall I delete current fXmin, fXmax ... and redefine them inside "boundaries"?

knoepfel:
@szhang17phys, you don't need to create a "boundaries" variable anywhere in the C++ code.  The code I alluded to before looks like:

```c++
PhotonGen::PhotonGen(fhicl::ParameterSet const& pset)
  : ...
{
  if (auto boundaries = pset.get_if_present<fhicl::ParameterSet>("boundaries")) {
    fXmin = boundaries->get<double>("Xmin");
    fXmax = boundaries->get<double>("Xmax");
    fYmin = boundaries->get<double>("Ymin");
    fYmax = boundaries->get<double>("Ymax");
    fZmin = boundaries->get<double>("Zmin");
    fZmax = boundaries->get<double>("Zmax");
  }
  else {
    art::ServiceHandle<geo::Geometry const> geo;
    auto const CryoBounds = geo->Cryostat().Boundaries();
    fXmin = CryoBounds.MinX();
    fXmax = CryoBounds.MaxX();
    fYmin = CryoBounds.MinY();
    fYmax = CryoBounds.MaxY();
    fZmin = CryoBounds.MinZ();
    fZmax = CryoBounds.MaxZ();
  }
}
```

where the `fXmin` (etc.) variables are used in both cases.

But the first `if` statement:

```c++
if (auto boundaries = pset.get_if_present<fhicl::ParameterSet>("boundaries")) {
 ...
}
```

is checking to see if the _**configuration file**_ has a "boundaries" parameter in the module's configuration:

```jsonnet
photonGen: {
  module_type: PhotonGen
  boundaries: {
    Xmin: ...
    Xmax: ...
    Ymin: ...
    ...
  }
}
```

Does that make sense?

szhang17phys:
Hello! Kyle. I made the modification based on your suggestion. The test on dunegpvm server looks fine. Hope this time it is good!

lgarren:
trigger build

knoepfel:
Thank you, @szhang17phys!  Looks good

lgarren:
trigger build with pull requests #47, uboone/ubreco#7

lgarren:
trigger build with #147, uboone/ubreco#7

lgarren:
Had to go with a local build of uboone, which is OK.

lgarren:
@LArSoft/level-2-managers please review and approve or comment

tomjunk (approved):
Approved.

Looks like a useful addition, and Kyle's comments have been addressed.

lgarren:
approve

PR #149, opened 2025-01-30
Changes necessary for Spack

knoepfel:
Should also work with UPS


lgarren:
trigger build

CMakeLists.txt
```diff
@@ -27,36 +27,34 @@ cet_cmake_module_directories(Modules BINARY)
...
 find_package(MARLEY REQUIRED EXPORT)
-find_package(PostgreSQL REQUIRED EXPORT)
 find_package(ROOT COMPONENTS Core EG GenVector Geom Gpad Hist MathCore Matrix Physics RIO RooFitCore Tree REQUIRED EXPORT)
-find_package(Range-v3 REQUIRED EXPORT)
-find_package(SQLite3 REQUIRED EXPORT)
-find_package(ifdhc REQUIRED EXPORT)
-find_package(log4cpp REQUIRED EXPORT)
+find_package(Range-v3 REQUIRED)
+find_package(SQLite3 REQUIRED)
+find_package(ifdhc REQUIRED)
```
lgarren, line 56: I think if you restore `find_package(ifdhc REQUIRED EXPORT)`, the DUNE build will succeed.

ups/product_deps
```diff
@@ -250,7 +250,6 @@ gdmldir	product_dir	gdml
 ####################################
 product		version		qual	flags		<table_format=2>
 larg4		v10_00_03	-
-larsoft_data	v1_02_02	-
```
lgarren, line 253: Why remove larsoft_data?  Pretty sure that files there are needed by bits of larsim.

CMakeLists.txt
knoepfel: You're right, @lgarren.  Because `dunesim` explicitly uses the `ifdh::ifdh` target, I think my preference is to create a `dunesim` PR with an explicit `find_package(ifdhc REQUIRED)` CMake command.

I'll play around with it and let you know.

ups/product_deps
knoepfel: Will check.

knoepfel: @lgarren, I've verified that `larsim` builds successfully standalone and all unit tests pass without `larsoft_data`.  If it turns out that a downstream package requires it, then we should add the dependency there.

lgarren: My concern is that some code may reference these files.  I doubt that the tests verify that, unfortunately.

knoepfel: > My concern is that some code may reference these files. I doubt that the tests verify that, unfortunately.

You're right--just checked.  Will revert this change.

CMakeLists.txt
lgarren: Agreeing that dunesim needs an explicit `find_package(ifdhc REQUIRED)`.  No change needed here.

knoepfel: Working on PR for dunesim.

knoepfel:
DUNE PR available: https://github.com/DUNE/dunesim/pull/85.

lgarren:
trigger build with pull requests #149, https://github.com/DUNE/dunesim/pull/85

lgarren:
approve

PR #152, opened 2025-04-01
MCRecoEdep fixes [1/2]

PetrilloAtWork:
The algorithm creating `sim::MCTrack` and `sim::MCShower` objects and data products builds upon "MC energy deposition" objects created by `MCRecoEdep` algorithm.
This algorithm was written for MicroBooNE with `sim::SimChannel` objects as input, and later extended to work also with `sim::SimEnergyDeposit` and `sim::SimEnergyDepositLite`. This extension is a separate algorithm, that was shaped after the former and not optimised.
This last implementation is the subject of this pull request.

It suffer from numerous bugs, including:
 * not checking from containment of the points on the planes when asking for the nearest wire to a deposit, which causes a flood of exceptions, typically gigabytes of log files and an according dramatic slow down;
 * because of a faulty check on the existence of a wire nearby¹, adding the same energy deposition (dE/dx) to multiple planes on different TPC;
 * potentially reporting the wrong plane when a deposit is sensed by a channel crossing different TPC;
 * another other minor bug on the determination of the detector volume (minor because the geometry of all experiment detectors using LArSoft do not trigger it).

I have fixed the bugs I have found, optimised the code a bit (nothing dramatic, percent improvement in speed), updated the comments which were outdated.
Now the code runs ×10 fast (because it does not waste its time throwing exceptions and printing messages for a query that is ultimately useless), does not force jobs to require tens of GB of disk space, and gives better results.

Did I write "better"? well, unfortunately there is not much testing on field, since I don't know effectively anybody actually using the dE/dx information I presumably fixed.
I have checked myself and was confirmed by ICARUS users that the changes do not break the algorithm. The number of reconstructed tracks and showers is almost identical, as the objects are as well, except for the dE/dx subject of the fixes.

In addition, I added output to `DumpMCParticles` module (I used it during debugging); **that change is bound to another in LArSoft/lardataalg#51, which has a twin PR**.

Note that despite fixing bugs on a quantity that no experiment seems to be using right now, this fix is still very important for the simulation workflow, because of the reduction of resources required to run the detector physics simulation ("GEANT4 stage").

--- 

**I ask for an exemption on the reformatting of `DumpMCParticle` module**: I have followed the style of the existing code, and since `format-code` reformats only the changes, it results in different styles for equivalent code. For example:
```cpp
    fhicl::Atom<std::string> OutputCategory{
      Name("OutputCategory"),
      Comment("name of the output stream (managed by the message facility)"),
      "DumpMCParticles" /* default value */
    };

    fhicl::Atom<bool> PrintStepPositions{
      Name("PrintStepPositions"),
      Comment("prints the position and time of each trajectory point"),
      true};
    fhicl::Atom<bool> PrintStepMomenta{Name("PrintStepMomenta"),
                                       Comment("prints the momentum at each trajectory point"),
                                       false};
    fhicl::Atom<bool> PrintStepEnergies{Name("PrintStepEnergies"),
                                        Comment("prints the total energy at each trajectory point"),
                                        false};

    fhicl::Atom<unsigned int> PointsPerLine{
      Name("PointsPerLine"),
      Comment("trajectory points printed per line (default: 2; 0 = skip them)"),
      2 /* default value */
    };

```
(`OutputCategory` and `PointsPerLine` show the original style).

--- 

¹ [`geo::PlaneGeo::NearewstWireID()`](https://code-doc.larsoft.org/docs/latest/html/classgeo_1_1PlaneGeo.html#a7d000aa0627101d8ec6a290068fc4406) has "undefined behaviour" (or, buggy behaviour) when used for a point not projecting on the plane.

### DIFF
--- a/larsim/MCSTReco/MCRecoEdep.cxx
+++ b/larsim/MCSTReco/MCRecoEdep.cxx
@@ -135,7 +135,9 @@ namespace sim {
     art::ServiceHandle<geo::Geometry const> geom;
 
     // Key map to identify a unique particle energy deposition point
-    std::map<std::pair<UniquePosition, unsigned int>, int> hit_index_m;
+    using TrackPoint_t = std::pair<UniquePosition, unsigned int>;
+    using HitIndices_t = std::map<TrackPoint_t, int>;
+    HitIndices_t hit_index_m;
 
     geo::PlaneDataContainer<std::size_t> const pindex = details::createPlaneIndexMap();
 
@@ -156,11 +158,7 @@ namespace sim {
       // next one.
       // Try first with the TPC from the previous point, if any; if it fails, go hunting
       if (!TPC || !TPC->ContainsPosition(mp)) TPC = geom->PositionToTPCptr(mp);
-      if (!TPC) {
-        // mf::LogWarning("SimDriftElectrons") << "step at " << mp << " is not within any TPC.";
-        // ... e.g. in the cathode
-        continue;
-      }
+      if (!TPC) continue;
 
       // make a collection of electrons for each plane
       for (geo::PlaneGeo const& plane : wireReadoutGeom.Iterate<geo::PlaneGeo>(TPC->ID())) {
@@ -182,9 +180,8 @@ namespace sim {
         // and all the planes in the TPC get the same charge contributions (barred geometric acceptance).
         // So: __GetEdepArray__[trackID][hitIndex].deps[planeIndex].energy/charge
         int const track_id = std::abs(sed.TrackID());
-        UniquePosition const pos{mp};
         std::vector<sim::MCEdep>& trackEDeps = __GetEdepArray__(track_id);
-        std::pair const key{pos, track_id};
+        TrackPoint_t const key{mp, track_id};
 
         // find the index of the hit the energy belongs to (possibly a new one):
         int hit_index = -1;
@@ -197,7 +194,7 @@ namespace sim {
         double const charge = sed.NumElectrons();
         if (hit_index < 0) {
           // This particle energy deposition is never recorded so far. Create a new Edep
-          trackEDeps.emplace_back(pos, planeid, pindex.size(), sed.Energy(), charge, planeNumber);
+          trackEDeps.emplace_back(mp, planeid, pindex.size(), sed.Energy(), charge, planeNumber);
         }
         else {
           // Append charge to the relevant edep (@ hit_index)
--- a/larsim/MCSTReco/MCRecoEdep.cxx
+++ b/larsim/MCSTReco/MCRecoEdep.cxx
@@ -141,6 +141,9 @@ namespace sim {
 
     if (_debug_mode)
       std::cout << "Processing " << sedArray.size() << " energy deposits..." << std::endl;
+    
+    geo::TPCGeo const* TPC = nullptr; // the TPC where the last deposit was seen
+    
     // Loop over channels
     for (size_t i = 0; i < sedArray.size(); ++i) {
 
@@ -162,7 +165,8 @@ namespace sim {
       // From the position in world coordinates, determine the cryostat and tpc. If
       // somehow the step is outside a tpc (e.g., cosmic rays in rock) just move on to the
       // next one.
-      geo::TPCGeo const* TPC = geom->PositionToTPCptr(mp);
+      
+      if (!TPC || !TPC->ContainsPosition(mp)) TPC = geom->PositionToTPCptr(mp);
       if (!TPC) {
         mf::LogWarning("SimDriftElectrons") << "step at " << mp << " is not within any TPC.";
         continue;
--- a/larsim/MCSTReco/MCRecoEdep.cxx
+++ b/larsim/MCSTReco/MCRecoEdep.cxx
@@ -146,6 +146,11 @@ namespace sim {
 
       // Get data to loop over
       auto const& sed = sedArray[i];
+      
+      mf::LogVerbatim log{ "MakeMCEdep" }; // FIXME DEBUG
+      log << "EDep=" << sed.Energy() << " Q=" << sed.NumElectrons()
+        << " at " << sed.MidPoint() << " t=" << sed.Time()
+        << " PDGID=" << sed.PdgCode() << " ID=" << sed.TrackID();
 
       // David Caratelli: much of the code below is taken from the module:
       // https://cdcvs.fnal.gov/redmine/projects/larsim/repository/revisions/develop/entry/larsim/ElectronDrift/SimDriftElectrons_module.cc
@@ -225,12 +230,17 @@ namespace sim {
           // This particle energy deposition is never recorded so far. Create a new Edep
           __GetEdepArray__(real_track_id)
             .emplace_back(pos, pid, pindex.size(), sed.Energy(), charge, channel_id);
+          log << "\n  plane #" << channel_id << " => pos=" << mp << " plane " << pid
+            << " (#" << channel_id << "/" << pindex.size() << ")"
+            " Q=" << charge << " E=" << sed.Energy();
         }
         else {
           // Append charge to the relevant edep (@ hit_index)
           MCEdep& edep = __GetEdepArray__(real_track_id).at(hit_index);
           edep.deps[channel_id].charge += charge;
           edep.deps[channel_id].energy += sed.Energy();
+          log << "\n  plane #" << channel_id << " => Q=" << edep.deps[channel_id].charge
+            << " E=" << edep.deps[channel_id].energy;
         }
       } // end looping over planes
     }   // end looping over SimEnergyDeposits
--- a/larsim/MCSTReco/MCRecoEdep.cxx
+++ b/larsim/MCSTReco/MCRecoEdep.cxx
@@ -147,10 +147,6 @@ namespace sim {
     // Loop over energy depositions
     for (sim::SimEnergyDeposit const& sed : sedArray) {
 
-      mf::LogVerbatim log{"MakeMCEdep"}; // FIXME DEBUG
-      log << "EDep=" << sed.Energy() << " Q=" << sed.NumElectrons() << " at " << sed.MidPoint()
-          << " t=" << sed.Time() << " PDGID=" << sed.PdgCode() << " ID=" << sed.TrackID();
-
       // David Caratelli: much of the code below is taken from the module:
       // https://cdcvs.fnal.gov/redmine/projects/larsim/repository/revisions/develop/entry/larsim/ElectronDrift/SimDriftElectrons_module.cc
 
@@ -202,15 +198,12 @@ namespace sim {
         if (hit_index < 0) {
           // This particle energy deposition is never recorded so far. Create a new Edep
           trackEDeps.emplace_back(pos, planeid, pindex.size(), sed.Energy(), charge, planeNumber);
-          log << "\n  plane #" << planeNumber << " => pos=" << mp << " plane " << planeid << " (#"
-              << planeNumber << "/" << pindex.size() << ") Q=" << charge << " E=" << sed.Energy();
         }
         else {
           // Append charge to the relevant edep (@ hit_index)
           MCEdep::deposit& dep = trackEDeps.at(hit_index).deps[planeNumber];
           dep.charge += charge;
           dep.energy += sed.Energy();
-          log << "\n  plane #" << planeNumber << " => Q=" << dep.charge << " E=" << dep.energy;
         }
       } // end looping over planes in TPC
     }   // end looping over SimEnergyDeposits
--- a/larsim/MCSTReco/MCRecoEdep.cxx
+++ b/larsim/MCSTReco/MCRecoEdep.cxx
@@ -162,34 +162,28 @@ namespace sim {
       // From the position in world coordinates, determine the cryostat and tpc. If
       // somehow the step is outside a tpc (e.g., cosmic rays in rock) just move on to the
       // next one.
-      unsigned int cryostat = 0;
-      try {
-        geom->PositionToCryostatID(mp);
-      }
-      catch (cet::exception& e) {
-        mf::LogWarning("SimDriftElectrons") << "step " // << energyDeposit << "\n"
-                                            << "cannot be found in a cryostat\n"
-                                            << e;
+      geo::TPCGeo const* TPC = geom->PositionToTPCptr(mp);
+      if (!TPC) {
+        mf::LogWarning("SimDriftElectrons") << "step at " << mp << " is not within any TPC.";
         continue;
       }
-      unsigned int tpc = 0;
-      try {
-        geom->PositionToTPCID(mp);
-      }
-      catch (cet::exception& e) {
-        mf::LogWarning("SimDriftElectrons") << "step " // << energyDeposit << "\n"
-                                            << "cannot be found in a TPC\n"
-                                            << e;
-        continue;
-      }
-      geo::TPCID const tpcid{cryostat, tpc};
-
+      
       // Define charge drift direction: driftcoordinate (x, y or z) and driftsign
       // (positive or negative). Also define coordinates perpendicular to drift direction.
 
       // make a collection of electrons for each plane
-      for (auto const& planeid : wireReadoutGeom.Iterate<geo::PlaneID>()) {
+      for (geo::PlaneGeo const& plane: wireReadoutGeom.Iterate<geo::PlaneGeo>(TPC->ID())) {
 
+        // require containment on the plane
+        if (!plane.isProjectionOnPlane(mp)) {
+          geo::PlaneGeo::WidthDepthProjection_t const deltaProj
+            = plane.DeltaFromPlane(plane.PointWidthDepthProjection(mp));
+          mf::LogVerbatim{ "SimDriftElectrons" }
+            << "Point " << mp << " is off " << plane.ID() << " by " << deltaProj.R() << " cm " << deltaProj;
+          continue;
+        }
+        geo::PlaneID const planeid = plane.ID();
+        
         // grab the nearest channel to the fDriftClusterPos position
         // David Caratelli, comment begin:
         // NOTE: the below code works only when the drift coordinate is indeed in x (i.e. 0th coordinate)
--- a/larsim/MCSTReco/MCRecoEdep.cxx
+++ b/larsim/MCSTReco/MCRecoEdep.cxx
@@ -18,8 +18,8 @@
 #include "MCRecoEdep.h"
 
 #include <iostream>
-#include <numeric> // std::iota()
 #include <map>
+#include <numeric> // std::iota()
 #include <utility>
 #include <vector>
 
@@ -30,7 +30,8 @@ namespace sim {
     {
       geo::WireReadoutGeom const& wireReadout = art::ServiceHandle<geo::WireReadout const>()->Get();
       geo::GeometryCore const& geom = *lar::providerFrom<geo::Geometry>();
-      geo::PlaneDataContainer<std::size_t> planeIndices{ geom.Ncryostats(), geom.MaxTPCs(), wireReadout.MaxPlanes() };
+      geo::PlaneDataContainer<std::size_t> planeIndices{
+        geom.Ncryostats(), geom.MaxTPCs(), wireReadout.MaxPlanes()};
       std::iota(planeIndices.begin(), planeIndices.end(), 0);
       return planeIndices;
     }
@@ -140,16 +141,15 @@ namespace sim {
 
     if (_debug_mode)
       std::cout << "Processing " << sedArray.size() << " energy deposits..." << std::endl;
-    
-    geo::TPCGeo const* TPC = nullptr; // the TPC where the last deposit was seen
-    
-    // Loop over energy depositions
-    for (sim::SimEnergyDeposit const& sed: sedArray) {
 
-      mf::LogVerbatim log{ "MakeMCEdep" }; // FIXME DEBUG
-      log << "EDep=" << sed.Energy() << " Q=" << sed.NumElectrons()
-        << " at " << sed.MidPoint() << " t=" << sed.Time()
-        << " PDGID=" << sed.PdgCode() << " ID=" << sed.TrackID();
+    geo::TPCGeo const* TPC = nullptr; // the TPC where the last deposit was seen
+
+    // Loop over energy depositions
+    for (sim::SimEnergyDeposit const& sed : sedArray) {
+
+      mf::LogVerbatim log{"MakeMCEdep"}; // FIXME DEBUG
+      log << "EDep=" << sed.Energy() << " Q=" << sed.NumElectrons() << " at " << sed.MidPoint()
+          << " t=" << sed.Time() << " PDGID=" << sed.PdgCode() << " ID=" << sed.TrackID();
 
       // David Caratelli: much of the code below is taken from the module:
       // https://cdcvs.fnal.gov/redmine/projects/larsim/repository/revisions/develop/entry/larsim/ElectronDrift/SimDriftElectrons_module.cc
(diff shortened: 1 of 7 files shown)


knoepfel:
@PetrilloAtWork, the formatting is not always perfect, but it is automatic.  Please adjust the code-formatting for this PR.  Thanks.

PetrilloAtWork:
Is it acceptable to reformat the entire file instead of only the changes, in order to achieve compliance _and_ consistency?

lgarren:
@PetrilloAtWork If you mean, should you run clang format on the entire file, then, yes, that is what is expected.  Usually only the changed code is affected.

PetrilloAtWork:
No, I actually meant to force the reformatting on the entire file rather than only on the changed lines, so that the resulting format is self-consistent.
That is at the price of making the process of tracking changes one step longer.

lgarren:
You can try.  In general, the answer would be no, but for this case you are welcome to make the attempt.  You will have to work within the limits of clang format.

PetrilloAtWork:
I am defeated.
I didn't appreciate that the weird format that Clang gives this code is not caused by parts that are not formatted because unchanged, but it's rather a (crookedly) consistent format: break after an open brace and indent, unless the lines are short, in which case align to the brace starting on its same line.
The result is horrible, but it is also consistent in a horrible way.

lgarren:
trigger build with pull requests #152, https://github.com/LArSoft/lardataalg/pull/51

larsim/MCSTReco/MCRecoEdep.cxx
```diff
@@ -137,106 +137,79 @@ namespace sim {
...
-      }
-      catch (cet::exception& e) {
-        mf::LogWarning("SimDriftElectrons") << "step " // << energyDeposit << "\n"
-                                            << "cannot be found in a TPC\n"
-                                            << e;
+      // Try first with the TPC from the previous point, if any; if it fails, go hunting
+      if (!TPC || !TPC->ContainsPosition(mp)) TPC = geom->PositionToTPCptr(mp);
+      if (!TPC) {
+        // mf::LogWarning("SimDriftElectrons") << "step at " << mp << " is not within any TPC.";
+        // ... e.g. in the cathode
```
greenc-FNAL: Are you removing a spurious warning here (then remove, not comment), or was this a temporary disablement (please restore)?

larsim/MCSTReco/MCRecoEdep.cxx
```diff
@@ -137,106 +137,79 @@ namespace sim {
...
-      unsigned int tpc = 0;
-      try {
-        geom->PositionToTPCID(mp);
-      }
-      catch (cet::exception& e) {
-        mf::LogWarning("SimDriftElectrons") << "step " // << energyDeposit << "\n"
-                                            << "cannot be found in a TPC\n"
-                                            << e;
+      // Try first with the TPC from the previous point, if any; if it fails, go hunting
+      if (!TPC || !TPC->ContainsPosition(mp)) TPC = geom->PositionToTPCptr(mp);
```
greenc-FNAL, line 160: Does `geom->PositionToTPCptr(mp)` have analogous error checking to the code you are removing here (and below)?

larsim/MCSTReco/MCRecoEdep.cxx
```diff
@@ -18,22 +19,21 @@
...
 #include <utility>
 #include <vector>
 
 namespace sim {
 
   namespace details {
-    std::map<geo::PlaneID, size_t> createPlaneIndexMap()
+    geo::PlaneDataContainer<std::size_t> createPlaneIndexMap()
     {
-      auto const& wireReadoutGeom = art::ServiceHandle<geo::WireReadout const>()->Get();
```
greenc-FNAL, line 29: In general, it is reasonable to use `auto` provided the variable's name is sufficiently explicatory.

larsim/MCSTReco/MCRecoEdep.cxx
```diff
@@ -137,106 +137,79 @@ namespace sim {
...
+
+        // for each track ID, we keep a list of "hits" (depositions at the exact same location);
+        // each hit tracks the energy and charge per plane (via a plane index);
+        // note which wire/channel senses the charge does not make any difference
+        // and all the planes in the TPC get the same charge contributions (barred geometric acceptance).
+        // So: __GetEdepArray__[trackID][hitIndex].deps[planeIndex].energy/charge
+        int const track_id = std::abs(sed.TrackID());
+        UniquePosition const pos{mp};
+        std::vector<sim::MCEdep>& trackEDeps = __GetEdepArray__(track_id);
+        std::pair const key{pos, track_id};
```
greenc-FNAL: This might be more robust against future changes if you were to have type aliases for the type of `hit_index_m` and its key type, and use them rather than relying on declaring the correct component types here (`UniquePosition` and `unsigned int`, at least right now).

larsim/MCSTReco/MCRecoEdep.h
```diff
@@ -45,47 +47,44 @@ namespace sim {
...
-    MCEdep(sim::UniquePosition p, geo::PlaneID pi, size_t num_planes, float e, float c, size_t id)
+    MCEdep(sim::UniquePosition p,
+           geo::PlaneID const& pi,
+           size_t num_planes,
+           float e,
+           float c,
+           size_t id)
       : pos(p), pid(pi), deps(num_planes)
     {
+      assert(id < num_planes);
```
greenc-FNAL: Maybe label (and document) this as a precondition?

larsim/MCSTReco/MCRecoEdep.h
```diff
@@ -45,47 +47,44 @@ namespace sim {
...
 
     MCEdep() = default;
 
-    MCEdep(sim::UniquePosition p, geo::PlaneID pi, size_t num_planes, float e, float c, size_t id)
+    MCEdep(sim::UniquePosition p,
+           geo::PlaneID const& pi,
+           size_t num_planes,
+           float e,
+           float c,
+           size_t id)
```
greenc-FNAL, line 85: Why does this struct hold a vector with `num_planes` uninitialized `deposit` objects, and then only initialize one of them?

larsim/MCSTReco/MCShowerRecoAlg.cxx
```diff
@@ -236,12 +237,11 @@ namespace sim {
...
-              shower_dep_dir[0] = edep.pos._x - mcshower[mcs_index].Start().X();
-              shower_dep_dir[1] = edep.pos._y - mcshower[mcs_index].Start().Y();
-              shower_dep_dir[2] = edep.pos._z - mcshower[mcs_index].Start().Z();
+              shower_dep_dir[0] = edep.pos.X() - mcshower[mcs_index].Start().X();
+              shower_dep_dir[1] = edep.pos.Y() - mcshower[mcs_index].Start().Y();
+              shower_dep_dir[2] = edep.pos.Z() - mcshower[mcs_index].Start().Z();
 
-              double dist = sqrt(pow(shower_dep_dir[0], 2) + pow(shower_dep_dir[1], 2) +
-                                 pow(shower_dep_dir[2], 2));
+              double dist = std::hypot(shower_dep_dir[0], shower_dep_dir[1], shower_dep_dir[2]);
```
greenc-FNAL: Protection against `dist==0`?

larsim/MCSTReco/MCShowerRecoAlg.cxx
```diff
@@ -286,12 +286,13 @@ namespace sim {
...
-          mom[0] = edep.pos._x - mcs_daughter_vtx[0];
-          mom[1] = edep.pos._y - mcs_daughter_vtx[1];
-          mom[2] = edep.pos._z - mcs_daughter_vtx[2];
+          mom[0] = edep.pos.X() - mcs_daughter_vtx[0];
+          mom[1] = edep.pos.Y() - mcs_daughter_vtx[1];
+          mom[2] = edep.pos.Z() - mcs_daughter_vtx[2];
 
           // Weight by energy (momentum)
-          double magnitude = sqrt(pow(mom.at(0), 2) + pow(mom.at(1), 2) + pow(mom.at(2), 2));
+          // double magnitude = sqrt(pow(mom.at(0), 2) + pow(mom.at(1), 2) + pow(mom.at(2), 2));
```
greenc-FNAL: Remove rather than comment old code.

larsim/MCSTReco/MCShowerRecoAlg.cxx
```diff
@@ -308,12 +309,13 @@ namespace sim {
...
           //Determine the direction of the shower right at the start point
           double E = 0;
           double N = 0;
-          if (sqrt(pow(edep.pos._x - mcs_daughter_vtx[0], 2) +
-                   pow(edep.pos._y - mcs_daughter_vtx[1], 2) +
-                   pow(edep.pos._z - mcs_daughter_vtx[2], 2)) < 2.4 &&
+          if (std::hypot(edep.pos.X() - mcs_daughter_vtx[0],
+                         edep.pos.Y() - mcs_daughter_vtx[1],
+                         edep.pos.Z() - mcs_daughter_vtx[2]) < 2.4 &&
               magnitude > 1.e-10) {
```
greenc-FNAL, line 320: Why `1.e-10` in particular? If this is not arbitrary, document it, and preferably define and use a well-named constant. If it is an arbitrary definition of "small," consider using `std::numeric_limits<double>::epsilon()`

larsim/MCSTReco/MCShowerRecoAlg.cxx
```diff
@@ -308,12 +309,13 @@ namespace sim {
...
+
           //Determine the direction of the shower right at the start point
           double E = 0;
           double N = 0;
-          if (sqrt(pow(edep.pos._x - mcs_daughter_vtx[0], 2) +
-                   pow(edep.pos._y - mcs_daughter_vtx[1], 2) +
-                   pow(edep.pos._z - mcs_daughter_vtx[2], 2)) < 2.4 &&
+          if (std::hypot(edep.pos.X() - mcs_daughter_vtx[0],
+                         edep.pos.Y() - mcs_daughter_vtx[1],
+                         edep.pos.Z() - mcs_daughter_vtx[2]) < 2.4 &&
```
greenc-FNAL, line 319: Please document this literal, and preferably define a well-named constant.

larsim/MCSTReco/MCShowerRecoAlg.cxx
```diff
@@ -362,10 +363,11 @@ namespace sim {
           // a*x + b*y + c*z + d = 0
           // where, a = dir_x, b = dir_y, c = dir_z, d = - (a*x_0+b*y_0+c*z_0)
           // then the *signed* distance of any point (x_1, y_1, z_1) from this plane is:
-          // D = (a*x_1 + b*y_1 + c*z_1 + d )/sqrt( pow(a,2) + pow(b,2) + pow(c,2))
+          // D = (a*x_1 + b*y_1 + c*z_1 + d )/hypot(a,b,c)
 
-          double p_mag = sqrt(pow(mcs_daughter_dir[0], 2) + pow(mcs_daughter_dir[1], 2) +
-                              pow(mcs_daughter_dir[2], 2));
+          // double p_mag = sqrt(pow(mcs_daughter_dir[0], 2) + pow(mcs_daughter_dir[1], 2) +
+          //                     pow(mcs_daughter_dir[2], 2));
```
greenc-FNAL: Remove rather than comment old code.

larsim/MCSTReco/MCShowerRecoAlg.cxx
```diff
@@ -378,11 +380,9 @@ namespace sim {
             continue;
           }
           //Radial Distance
-          if ((a * edep.pos._x + b * edep.pos._y + c * edep.pos._z + d) /
-                  sqrt(pow(a, 2) + pow(b, 2) + pow(c, 2)) <
+          if ((a * edep.pos.X() + b * edep.pos.Y() + c * edep.pos.Z() + d) / std::hypot(a, b, c) <
```
greenc-FNAL: Check for zero devisor.

larsim/MCSTReco/MCShowerRecoAlg.cxx
```diff
@@ -378,11 +380,9 @@ namespace sim {
...
                 2.4 &&
-              (a * edep.pos._x + b * edep.pos._y + c * edep.pos._z + d) /
-                  sqrt(pow(a, 2) + pow(b, 2) + pow(c, 2)) >
+              (a * edep.pos.X() + b * edep.pos.Y() + c * edep.pos.Z() + d) / std::hypot(a, b, c) >
```
greenc-FNAL: Check for zero devisor, or use a temporary (verified non-zero) variable in both places.

larsim/MCSTReco/MCShowerRecoAlg.cxx
```diff
@@ -378,11 +380,9 @@ namespace sim {
             continue;
           }
           //Radial Distance
-          if ((a * edep.pos._x + b * edep.pos._y + c * edep.pos._z + d) /
-                  sqrt(pow(a, 2) + pow(b, 2) + pow(c, 2)) <
+          if ((a * edep.pos.X() + b * edep.pos.Y() + c * edep.pos.Z() + d) / std::hypot(a, b, c) <
                 2.4 &&
```
greenc-FNAL: See above re hardwired literals. This should be a single point of maintenance.

larsim/MCSTReco/MCShowerRecoAlg.cxx
```diff
@@ -378,11 +380,9 @@ namespace sim {
...
-              (a * edep.pos._x + b * edep.pos._y + c * edep.pos._z + d) /
-                  sqrt(pow(a, 2) + pow(b, 2) + pow(c, 2)) >
+              (a * edep.pos.X() + b * edep.pos.Y() + c * edep.pos.Z() + d) / std::hypot(a, b, c) >
                 0) {
```
greenc-FNAL: See above re hardwired literals.

larsim/MCSTReco/MCTrackRecoAlg.cxx
```diff
@@ -184,11 +186,9 @@ namespace sim {
           // Add in a voxel before and after to account for MCSteps
           // the line distance allows for 1mm GEANT multiple columb scattering correction,
           // small compared to average MCStep-to-MCStep distance
-          if ((a * edep.pos._x + b * edep.pos._y + c * edep.pos._z + d) /
-                  sqrt(pow(a, 2) + pow(b, 2) + pow(c, 2)) <=
+          if ((a * edep.pos.X() + b * edep.pos.Y() + c * edep.pos.Z() + d) / std::hypot(a, b, c) <=
                 dist + 0.03 &&
```
greenc-FNAL, line 190: See above re literals.

larsim/MCSTReco/MCTrackRecoAlg.cxx
```diff
@@ -184,11 +186,9 @@ namespace sim {
...
-              (a * edep.pos._x + b * edep.pos._y + c * edep.pos._z + d) /
-                  sqrt(pow(a, 2) + pow(b, 2) + pow(c, 2)) >=
+              (a * edep.pos.X() + b * edep.pos.Y() + c * edep.pos.Z() + d) / std::hypot(a, b, c) >=
                 0 - 0.03 &&
               LineDist < 0.1) {
```
greenc-FNAL, line 193: See above re literals.

greenc-FNAL (changes requested):

larsim/MCSTReco/MCRecoEdep.cxx
PetrilloAtWork: My rule of thumb is that (the documentation of) the type can be easily reached with average knowledge of C++. For example, `std::make_unique<DataProduct>()` spells out the type, `myVector.begin()` is well known, etc.
Here it's subtle since the type mentioned is `geo::WireReadout` and the result is `geo::WireReadoutGeom`, similar but different (and at this point quite a new name, for what it's worth in the long term).
However, if you feel strongly, I can change it: just let me know in this thread.

larsim/MCSTReco/MCRecoEdep.cxx
PetrilloAtWork: Yes: the old code can react to the points not being in the volumes they cover (cryostats and TPC) by throwing an exception. The new code reacts to the same condition, by returning a special value.

larsim/MCSTReco/MCRecoEdep.cxx
PetrilloAtWork: Missed... done now.

larsim/MCSTReco/MCRecoEdep.cxx
PetrilloAtWork: I hadn't bothered because it's all in the same function.
However, after your suggestion I gave a try, and it does not look bad (it ends up hiding `UniquePosition`, which is basically a sortable 3D point but it does not say).

larsim/MCSTReco/MCRecoEdep.h
PetrilloAtWork: Here and in some of the following questions, I am oblivious to many of the why's of this code, so I just adventure into guesses.
This object was obviously not designed to record which planes are used and which are not, and assumes that the user will know. I don't fully understand the downstream code, so I can't even assure that the aforementioned user code  takes safe steps to avoid using uninitialised data.
However, in an excess of caution, I have added what I hope to be sensible initialisation values to the `deposit` class.

larsim/MCSTReco/MCRecoEdep.h
PetrilloAtWork: How do you suggest to do both?

larsim/MCSTReco/MCShowerRecoAlg.cxx
PetrilloAtWork: It would be nice, but I have no idea what a sensible value for `v` would be in that case, or what is the sensible action to take when that happens.

I tried, but I am not confident that is the best approach.

larsim/MCSTReco/MCShowerRecoAlg.cxx
PetrilloAtWork: Agreed.
I think I left it as it was one of the two places where my changes actually changed the results, for unknown reasons.

larsim/MCSTReco/MCShowerRecoAlg.cxx
PetrilloAtWork: Unfortunately I have no idea where this number comes from. I can guess it's a length (since it bridges the value between a energy sum and a dE/dx), but I am not going to write anything that might turn out wrong.

By the way, I bet it depends on the pitch between the detector wires.

larsim/MCSTReco/MCShowerRecoAlg.cxx
PetrilloAtWork: I guess it's the second, but I am not sure.
Also, I am not sure epsilon() is the right choice, considering that the result might be some function of ε which makes it larger (for example the sum of three of them, √3·ε).

larsim/MCSTReco/MCShowerRecoAlg.cxx
PetrilloAtWork: ... and here is the other place.

larsim/MCSTReco/MCShowerRecoAlg.cxx
PetrilloAtWork: Yes — and same reason for me not doing it.

larsim/MCSTReco/MCShowerRecoAlg.cxx
PetrilloAtWork: Fine for the temporary value. Although there are plenty of these things in this code, and I didn't intend to fix it (my intention was just to address `pow(x, 2)` for... personal reasons? :sweat_smile:).

About protection: it turns out to be already protected, and in fact `std::hypot(a, b, c)` should be always `1` (it's the unit vector of `mcs_daughter_dir`, which is in turn already "protected").

And no, I am not going to move all that `mcs_daughter_dir` code out of the ~loop~ _two loops_, because it's a rabbit hole.

larsim/MCSTReco/MCShowerRecoAlg.cxx
PetrilloAtWork: Agree, but won't fix.

larsim/MCSTReco/MCTrackRecoAlg.cxx
PetrilloAtWork: Agree, but won't fix.

larsim/MCSTReco/MCTrackRecoAlg.cxx
PetrilloAtWork: Agree, but won't fix.

PetrilloAtWork:
I have gone through the recommendations.
I agree with all of them but one, yet I did not act on many of them, because often my understanding of the code is not enough to perform meaningful changes.
Just to make an example, there are several `2.4` "magic" values in the code, and I can't even tell if they are all bound together or not.
The essence of this PR is bug fixes in `MCRecoEdep`, plus a change in `sqrt(pow(x, 2)+...)` because I am physically unable to stand it.

larsim/MCSTReco/MCRecoEdep.cxx
greenc-FNAL: Since C++17, one can use [`std::optional`](https://en.cppreference.com/w/cpp/utility/optional) to indicate that a function cannot return a valid value, rather than returning a "special" value that one should interpret as invalid. A simple conversion to bool can verify validity:
```
if (opt_val) { // Do stuff with *opt_val)`
  ...
}
```

larsim/MCSTReco/MCRecoEdep.cxx
greenc-FNAL: Your rule of thumb seems perfectly resonable.

larsim/MCSTReco/MCRecoEdep.h
greenc-FNAL: I would add an inline comment (e.g.) `// prerequisite`, and mention the condition in the declaration's Doxygen documentation for the `id` parameter, if it exists.

larsim/MCSTReco/MCShowerRecoAlg.cxx
greenc-FNAL: Already protected is good, but in that case you might want to document the condition as a prerequisite checked via `assert()` to cover you against someone removing the check from the preceding code.

larsim/MCSTReco/MCRecoEdep.h
greenc-FNAL: I would suggest marking with a `TODO`: it's perfectly reasonable to avoid changing code one doesn't understand, but the `TODO` might trigger a neuron in someone who _does_ understand the code well enough to determine whether it is wrong, and fix it.

larsim/MCSTReco/MCShowerRecoAlg.cxx
greenc-FNAL: See above re suggested `TODO`.

larsim/MCSTReco/MCShowerRecoAlg.cxx
greenc-FNAL: See above re suggested TODO.

larsim/MCSTReco/MCTrackRecoAlg.cxx
greenc-FNAL: See above re suggested TODO.

larsim/MCSTReco/MCTrackRecoAlg.cxx
greenc-FNAL: See above re suggested TODO.

larsim/MCSTReco/MCRecoEdep.cxx
PetrilloAtWork: Do you refer to `geo::GeometryCore::PositionToTPCptr()`?
Besides your general point, that interface predates the adoption of C++17 in LArSoft. Bare pointers expose roughly the same interface in terms of validity check and payload access, and I would think checking the validity of a pointer is routine. `std::optional` might have been more useful in `geo::GeometryCore::PositionToTPC()`, if it had supported references, which to my understanding it does not (yet).

larsim/MCSTReco/MCRecoEdep.h
PetrilloAtWork: Added the comment. There is no Doxygen documentation for the class, so I am not introducing it in there.

larsim/MCSTReco/MCRecoEdep.h
PetrilloAtWork: I have added a `@todo`, although I don't hold my breath as I had hard time finding people with knowledge about the internals of this algorithm.

PetrilloAtWork:
I have added comments (Doxygen and inline) to reflect the recommendations where possible.

larsim/MCSTReco/MCTrackRecoAlg.cxx
PetrilloAtWork: I have added a `@todo` in the class Doxygen documentation, since this is an issue across its code.

larsim/MCSTReco/MCTrackRecoAlg.cxx
PetrilloAtWork: :point_up:

larsim/MCSTReco/MCRecoEdep.cxx
greenc-FNAL: You are correct that `std::optional` does not support references, and that `std::optional<T*>` is of limited advantage with respect to a returned null pointer. It does, however, provide annotation for the likelihood of an invalid result which may be desirable under some circumstances. In this particular case though, I withdraw my comment.

lgarren:
trigger build with pull requests https://github.com/LArSoft/larsim/pull/152, https://github.com/LArSoft/lardataalg/pull/51

greenc-FNAL (approved):

absolution1:
approve

lgarren:
approve

PR #155, opened 2025-06-12
IonAndScint:GapChargeTransport

dahendric:
Added a tool recovering the charge deposited in the gaps between CRPs (in DUNE-FD-VD) into the IonAndScint module.

### DIFF
--- a/larsim/IonizationScintillation/IGapChargeTransport.hh
+++ b/larsim/IonizationScintillation/IGapChargeTransport.hh
@@ -1,0 +1,21 @@
+#ifndef LARSIM_IGAPCHARGETRANSPORTTOOL_HH
+#define LARSIM_IGAPCHARGETRANSPORTTOOL_HH
+
+#include "larcorealg/Geometry/geo_vectors_utils.h"
+#include <utility>
+
+namespace gap {
+  class IGapChargeTransport {
+  public:
+    virtual ~IGapChargeTransport() = default;
+
+    /// Whether to apply gap charge transport function
+    virtual bool EnableGapChargeTransport() const = 0;
+
+    /// Return shifted position and number of moved electrons
+    virtual std::pair<geo::Point_t, int> GetOffset(
+      double x, double y, double z, int n) const = 0;
+  };
+}
+
+#endif
--- a/larsim/IonizationScintillation/IGapChargeTransport.hh
+++ b/larsim/IonizationScintillation/IGapChargeTransport.hh
@@ -5,7 +5,7 @@
 #include <utility>
 
 namespace gap {
-  class IGapChargeTransport {
+  class IGapChargeTransport { // used to recover charge from the gap between CRPs
   public:
     virtual ~IGapChargeTransport() = default;
 
--- a/larsim/IonizationScintillation/IGapChargeTransport.hh
+++ b/larsim/IonizationScintillation/IGapChargeTransport.hh
@@ -13,8 +13,7 @@ namespace gap {
     virtual bool EnableGapChargeTransport() const = 0;
 
     /// Return shifted position and number of moved electrons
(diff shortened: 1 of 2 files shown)


larsim/IonizationScintillation/IonAndScint_module.cc
```diff
@@ -239,17 +264,32 @@ namespace larg4 {
...
-                              startPos_tmp,
-                              endPos_tmp,
-                              startTime_tmp,
-                              endTime_tmp,
-                              trackID_tmp,
-                              pdgCode_tmp,
-                              origTrackID_tmp);
+        //----------gap charge-----------------------
+        if (fGapTool) {
+          if (edeps.provenance()->productInstanceName() == "LArG4DetectorServicevolEnclosureTPC") {
```
lpaulucc: Would this be better put as a fhicl parameter instead of a hard coded value?

larsim/IonizationScintillation/IonAndScint_module.cc
```diff
@@ -118,6 +137,11 @@ namespace larg4 {
 
     produces<std::vector<sim::SimEnergyDeposit>>();
     if (fSavePriorSCE) produces<std::vector<sim::SimEnergyDeposit>>("priorSCE");
+
+    if (fGapTool) { std::cout << "fGapTool is configured" << std::endl; }
+    else {
+      //std::cout << "fGapTool is nullptr (tool not configured)" << std::endl;
+    }
```
absolution1: This looks like it can be dropped

larsim/IonizationScintillation/IonAndScint_module.cc
```diff
@@ -136,6 +160,7 @@ namespace larg4 {
       fISAlg = std::make_unique<ISCalcNESTLAr>(fEngine);
     else
       mf::LogWarning("IonAndScint") << "No ISCalculation set, this can't be good.";
+    if (fGapTool) std::cout << "Tool on, begin job" << std::endl;
```
absolution1: the couts should use the messaging service.  There's an example of how to use the messaging service once line above this.

larsim/IonizationScintillation/IonAndScint_module.cc
absolution1: Agreed, different LArTPCs will use different naming conventions so it would be best to make this run-configurable (through fcl)

absolution1 (commented):
Hi @dahendric 
Thank you for submitting this PR.
It looks like a very nice and clean addition.  I only suggest a few minor changes.  Most could be ignored if you wish, but I think the volume name setting should be changed to be read in through fcl (as per @lpaulucc's suggestion)

larsim/IonizationScintillation/IGapChargeTransport.hh
```diff
@@ -0,0 +1,20 @@
...
+#include <utility>
+
+namespace gap {
+  class IGapChargeTransport { // used to recover charge from the gap between CRPs
+  public:
+    virtual ~IGapChargeTransport() = default;
+
+    /// Whether to apply gap charge transport function
+    virtual bool EnableGapChargeTransport() const = 0;
+
```
knoepfel: No need to keep the `EnableGapChargeTransport()` virtual function.

dahendric:
@lpaulucc, @absolution1, @knoepfel thank you for your comments! I've updated the code, the volume name is no longer hardcoded (I still use it in dunesim though to prevent it running with a typo in the name received  from the fhicl), couts replaced with messaging service and .hh cleaned

lpaulucc (approved):
Thank you for the adjustments @dahendric. It looks good to me.

knoepfel:
trigger build

lgarren:
approve

PR #156, opened 2025-06-25
Fix SimPhoton time rounding

mvicenzi:
This PR fixes the incorrect rounding of photon times when producing `SimPhotons` collections.

Both `PDFastSimPAR` and `PDFastSimPVS` modules currently round the photon times to the closest integer when preparing their `SimPhotons` collections. While the rounding is needed for the `SimPhotonsLite` approach -- time is used as index for the vector -- this should not be done when saving each individual `sim::OnePhoton` time.

### DIFF
--- a/larsim/PhotonPropagation/PDFastSimPAR_module.cc
+++ b/larsim/PhotonPropagation/PDFastSimPAR_module.cc
@@ -563,8 +563,7 @@ namespace phot {
                 // calculates the time at which the photon was produced
                 double dtime = edepi.StartT() + fScintTime->fastScintTime();
                 if (fIncludePropTime) dtime += transport_time[i];
-                int time = static_cast<int>(std::round(dtime));
-                photon.Time = time;
+                photon.Time = dtime;
                 if (Reflected)
                   ref_photcol[channel].insert(ref_photcol[channel].end(), 1, photon);
                 else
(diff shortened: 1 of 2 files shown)


lpaulucc (commented):
@mvicenzi Thank you for identifying this. It might have been on purpose as the size of SimPhotons is already substantially larger than SimPhotonsLite. Do you have a user case where this change is needed as in the end the times will be converted to ticks during digitization? Do you also have an estimate on performance with this change? It is also worth pointing out that the OpBackTracker is still receiving time instead of dtime.

mvicenzi:
We don’t have a specific use case where we see a difference yet, but it could become important when using a response template with sub-nanosecond ticks. This is something one might consider when dealing with a steep rising edge.
I did some checks on the performance side. The `sim::OnePhoton` time is already stored as a `float`, so the size of the data product in memory doesn't change. However, it's true that the compression in the ROOT file for this branch becomes slightly less efficient. I see the file size increasing by about 3% when I disable the rounding.

Not sure what to do for `OpDetBacktracker`, I don't have experince using it... maybe it's okay to leave the time rounded to keep it "lite" and match exactly what ends up in `SimPhotonsLite`?

knoepfel:
@mvicenzi, thank you for the PR.  Although this PR cleans up the code, it may have consequences (e.g. narrowing conversions) that are undesirable in other contexts.  We will reach out to other photon folks for their input.  Stay tuned.

lgarren:
approve

approved by software coordinators

PR #158, opened 2025-07-17
Update QAlpha from 0.72 to 0.33 for alpha simulation

RalAngelo:
## Context

This pull request updates the `Qalpha` parameter from 0.72 to 0.33 in alignment with the observed light yield in alpha particle simulations for ProtoDUNE-VD.

## Justification

We observed the following results when simulating a 5.36 MeV alpha particle:

- With Qalpha = 0.72:
  - Photons produced = 195,144
  - Photons/MeV = 36,407

- With Qalpha = 0.33:
  - Photons produced = 89,422
  - Photons/MeV = 16,683

## Acknowledgments

This change was discussed with the ProtoDUNE DRA Photon Detection System team, led by:

- Jose Alfonso Soto Oton (j.soto@cern.ch)
- Laura Paulucci (laura.paulucci@ufabc.edu.br)

Laura Paulucci has confirmed approval for this change and suggested that it be submitted via pull request to LArSoft for review.

## Author

- Angelo Ralaikoto (ralaikotoangelo@gmail.com)
- Project: "Alpha particle simulation studies - ProtoDUNE-VD"

### DIFF
--- a/larsim/Simulation/simulationservices.fcl
+++ b/larsim/Simulation/simulationservices.fcl
@@ -65,7 +65,7 @@ standard_largeantparameters:
 
   #* alpha particle quenching factor
   #* Doke et al. Jpn. J. Appl. Phys. Vol. 41 (2002) pp. 1538
- QAlpha: 0.72
+ QAlpha: 0.33
 
   #* ion+excitation work function in eV
   #* https://doi.org/10.1016/0168-9002(90)90011-T


lpaulucc (approved):
This PR is fixing the light yield obtained for the alpha particle in the correlated model, as pointed out by Angelo. The alpha particle light yield deserves a separate treatment from other particles given the high ionization density along its track. Nevertheless such treatment is still missing. In order to make the light yield given by the Correlated model compatible with the one given by ScintByParticleType, the alpha quenching factor has to be smaller than the one given in reference Doke et al., Jpn. J. Appl. Phys. Vol. 41 (2002) pp. 1538–1545.

ggamezdiego:
Hi @lpaulucc, regarding the light yield value used in ScintByParticleType for alpha particles, we have tried to reproduce it and believe it is not correct. In the file [larproperties](https://github.com/LArSoft/lardataalg/blob/6d6c7b65f0cd8c495a0043e697aa1a8a258214ca/lardataalg/DetectorInfo/larproperties.fcl).fcl, the value used is AlphaScintYield: 16800. This number can be reproduced with the following calculation: 40000 (supposedly LY_max at 0 electric field) × 0.7 (quenching factor) × 0.6 (reduction at 500 V/cm) = 16800 photons/MeV. However, the correct value, if I understand the Doke paper you referenced above, should be: 51282 (LY_max at 0 electric field → ΔE / W_ph) × 0.72 (quenching factor) × 0.6 (reduction at 500 V/cm) = 22154 photons/MeV.
Therefor, I fully agree that we need to review/reduce the number of photons generated by alpha particles, but I would say that the attenuation factor relative to LY_max should be 0.72 × 0.6 = 0.43. We would need to verify whether, for the dE/dx of alpha particles, the current model predicts the LY_max value (before applying any type of quenching), and if it does not, include an additional correction factor to account for that.

lpaulucc:
@ggamezdiego I confirm that LArQL is predicting around 50k photon/MeV @500 V/cm without corrections in the LArQL model.

knoepfel:
@ggamezdiego and @lpaulucc, is this PR ready for approval, or are there additional issues that need to be sorted out?

lpaulucc:
@RalAngelo Could you update the alpha quenching to 0.43 as suggested by @ggamezdiego?

lpaulucc:
Instead of updating this PR, another one was created (https://github.com/LArSoft/larsim/pull/160). I will approve the one there but I am not sure how this is better handled.

lgarren:
It would have been fine to update this PR to preserve the conversation.  We will close this PR.

PR #157, opened 2025-07-16
Updates to semi-analytical light simulation: optical path tool

pgreen135:
This PR adds a tool "OpticalPath" for identifying which photon detectors are visible depending on where the light emission occurs for detectors where there are multiple optically isolated volumes. This replaces the placeholder solution discussed in relation to https://github.com/LArSoft/larsim/pull/145 in January. 

The default behaviour of the tool is that all optical detectors are visible for any emission point. This case works in DUNE-HD 1x2x6, DUNE VD, protoDUNE VD and MicroBooNE.

Custom cases are implemented for: 
- SBND: https://github.com/SBNSoftware/sbndcode/pull/756
(Note: SBND is currently several releases behind LArSoft, so the CI checks will fail; but this has been tested locally using cherry-picked build) 
- DUNE-HD 10kt, protoDUNE-HD: https://github.com/DUNE/dunesw/pull/186 and https://github.com/DUNE/duneopdet/pull/104

Also requires: https://github.com/LArSoft/larwirecell/pull/58 

The PR also implements two minor/maintenance changes to the models:
- adds a safeguard for vuv timing parameterisations going out of range, to avoid potential segfault
- adds a special case to apply border corrections in vertical direction only for use in DUNE-HD 10kt

PR #159, opened 2025-08-07
PDFastSimPAR can now be configured to restrict the TPCs in which edeps can produce photons

calcuttj:
Mainly this is to help PDHD's memory utilization.

Configured using a sequence of integers representing TPC numbers. In the case of an empty sequence, the module produces photons from all edeps/TPCs. Default configuration is an empty sequence.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,7 +9,7 @@
 cmake_minimum_required(VERSION 3.20 FATAL_ERROR)
 
 find_package(cetmodules 3.24.00 REQUIRED)
-project(larsim VERSION 10.06.03 LANGUAGES CXX)
+project(larsim VERSION 10.07.00 LANGUAGES CXX)
 
 include(CetCMakeEnv)
 cet_cmake_env()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,7 +9,7 @@
 cmake_minimum_required(VERSION 3.20 FATAL_ERROR)
 
 find_package(cetmodules 3.24.00 REQUIRED)
-project(larsim VERSION 10.06.02 LANGUAGES CXX)
+project(larsim VERSION 10.06.03 LANGUAGES CXX)
 
 include(CetCMakeEnv)
 cet_cmake_env()
--- a/larsim/PhotonPropagation/CMakeLists.txt
+++ b/larsim/PhotonPropagation/CMakeLists.txt
@@ -1,5 +1,6 @@
 add_subdirectory(LibraryMappingTools)
 add_subdirectory(ScintTimeTools)
+add_subdirectory(OpticalPathTools)
 
 cet_make_library(LIBRARY_NAME PhotonLibraryInterface INTERFACE
   SOURCE IPhotonLibrary.h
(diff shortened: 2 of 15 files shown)


knoepfel:
trigger build

lpaulucc:
Hi Jake,
How does this PR compares with this other one https://github.com/LArSoft/larsim/pull/157 ? Do you wish to be even more restrictive than the conditions foreseen in this PR (based on PD-Edep point distance and an opaque cathode)?

calcuttj:
Hi Laura, this is a complementary change to that PR and is more or less an extension of the mechanism for choosing to simulate in/out of the active volume

tomjunk:
Sounds fine, as long as the default behavior is not changed.  We can of course shoot ourselves in the foot with this -- effectively turning off a fixed portion of the detector.

tomjunk:
Approved

pgreen135:
I also see no issue with this, provided that:
a). It doesn't conflict with https://github.com/LArSoft/larsim/pull/157, which it doesn't look like it does. If I'm understanding the code correctly its similar logic to the inside/outside active volume, rather impacting which PDs are visible from which TPC. So for a given edep, if its in a TPC volume that is within the list; all PDs will be visible even those in other TPCs unless breaching the rule configured in the tool in https://github.com/LArSoft/larsim/pull/157 / the old isOpDetInSameTPC function --- @calcuttj correct me if that is not the case, as that would be an issue.
b).  the default behaviour is to simulate everything such that SBND, uBooNE, ICARUS, DUNE-HD and DUNE-VD are not impacted unless they choose to use this feature

lpaulucc:
Thank you for your answer, Jake. If I understand correctly, this is more general as you choose TPCs not to be processed at all instead of based on the position of Edep with respect to the OpChannel number. I am fine with it but this should be used with much care.

calcuttj:
@pgreen135 
> So for a given edep, if its in a TPC volume that is within the list; all PDs will be visible even those in other TPCs unless breaching the rule configured in the tool in #157 / the old isOpDetInSameTPC function --- @calcuttj correct me if that is not the case, as that would be an issue. 

This is correct. So it won't be an issue. It's just whether or not the edep itself is considered, and is not relative to the OpDet in question

> b). the default behaviour is to simulate everything such that SBND, uBooNE, ICARUS, DUNE-HD and DUNE-VD are not impacted unless they choose to use this feature

Correct

calcuttj:
Is anything needed to get this merged in? I'd really like to get this in next week's release or even potentially a patch release if that's possible

lpaulucc (approved):

lgarren:
@calcuttj There was enough confusion with #157 that it seemed best not to merge this in the same release.

lgarren:
@calcuttj This no longer merges as intended with develop.  My attempts to find a way to merge it cleanly have failed.  Would you rebase and make sure it is as intended?  Thanks!

calcuttj:
Rebased against develop and got it to build + tested that it works as intended

lgarren:
Thanks @calcuttj !

lgarren:
trigger build

lgarren:
approve

calcuttj:
Thank you Lynn

lgarren:
Huh?  Who is asking?

gartung:
Probably added by one of the Jenkins plugins I was testing.

PR #161, opened 2025-09-05
Remove use of art's `art::Event::getView`

knoepfel:
Only containers of type `std::vector<sim::SimChannel>` are written to disk. Using an art view is therefore unnecessary and, in anticipation of the Phlex framework (which will not support views), this PR changes to using the more conventional `evt.getProduct<std::vector<sim::SimChannel>>(...)`.


lgarren:
trigger build

lgarren:
trigger build

lgarren:
approve

PR #162, opened 2025-10-23
Fix MutliPart Generator random Engine

bear-is-asleep:
Update multipart generator to have engines run without segmentation faults. Segmentation faults occur in SBND without this patch.

@LauPM can you check this?

### DIFF
--- a/larsim/EventGenerator/MultiPart/MultiPartRain_module.cc
+++ b/larsim/EventGenerator/MultiPart/MultiPartRain_module.cc
@@ -134,16 +134,20 @@ MultiPartRain::MultiPartRain(fhicl::ParameterSet const& p)
 
   // Random engine initialization
   fFlatRandom = std::make_unique<CLHEP::RandFlat>(fFlatEngine, 0, 1);
-  fNormalRandom = std::make_unique<CLHEP::RandGauss>(fFlatEngine);
-  // x^2 distribution for fCosmicAngleRandom
-  // will be used to draw cos(theta)
-  const int nbins(100);
-  double parent[nbins];
-  for (size_t idx = 0; idx < nbins; ++idx) {
-    //parent[idx] = cet::square(cos(((float) idx)/nbins * util::pi()));
-    parent[idx] = cet::square(((float)idx) / nbins); //
+  if (_cosmic_distribution) {
+    // x^2 distribution for fCosmicAngleRandom
+    // will be used to draw cos(theta)
+    const int nbins(100);
+    double parent[nbins];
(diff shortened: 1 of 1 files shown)


knoepfel:
@bear-is-asleep, the reason for the segfault is that the variable `bool _cosmic_distribution`[ is initialized](https://github.com/LArSoft/larsim/blob/1daf32e0e5e587de6b340cb1a7a081934b018b77/larsim/EventGenerator/MultiPart/MultiPartRain_module.cc#L163) *after* [it is used for the first time](https://github.com/LArSoft/larsim/blob/1daf32e0e5e587de6b340cb1a7a081934b018b77/larsim/EventGenerator/MultiPart/MultiPartRain_module.cc#L137).

The best fix would be to initialize `_cosmic_distribution` in the initialization list of the module:

```c++
MultiPartRain::MultiPartRain(fhicl::ParameterSet const& p)
  : EDProducer(p)
  , fFlatEngine(art::ServiceHandle<rndm::NuRandomService>()->registerAndSeedEngine(
      createEngine(0, "HepJamesRandom", "GenRain"),
      "HepJamesRandom",
      "GenRain"))
  , _cosmic_distribution{p.get<bool>("CosmicDistribution", false)} // <== Initialize before use
{
 // ... a bunch of code that can now use _cosmic_distribution
}
```

bear-is-asleep:
> ```c++
> , _cosmic_distribution{p.get<bool>("CosmicDistribution", false)}
> ```

It successfully builds locally after the most recent commit. Thank you for the suggestion!

knoepfel:
> > ```c++
> > , _cosmic_distribution{p.get<bool>("CosmicDistribution", false)}
> > ```
> 
> It successfully builds locally after the most recent commit. Thank you for the suggestion!

No problem.  Actually, that one change should be sufficient to resolve the segfault—you shouldn't need the other commits.  Can you adjust this PR to contain only the change in initializing the Boolean variable?

bear-is-asleep:
> ge in initializing the Boolean variable?



> > > ```c++
> > > , _cosmic_distribution{p.get<bool>("CosmicDistribution", false)}
> > > ```
> > 
> > 
> > It successfully builds locally after the most recent commit. Thank you for the suggestion!
> 
> No problem. Actually, that one change should be sufficient to resolve the segfault—you shouldn't need the other commits. Can you adjust this PR to contain only the change in initializing the Boolean variable?

Ah I got it, just updated the PR to only use the boolean. I checked and it locally builds and runs.

lgarren:
trigger build

lgarren:
approve

PR #164, opened 2025-12-06
replace the infinite while loop in multipart rain module

nvpm-lab:
While running the ProtoDUNE-VD MPVMP generator, the infinite loop in `MultiPartRain_module.cc` was leading to segmentation faults. Constraining the generated direction to a single octant instead of the full sphere resolved the issue.

### DIFF
--- a/larsim/EventGenerator/MultiPart/MultiPartRain_module.cc
+++ b/larsim/EventGenerator/MultiPart/MultiPartRain_module.cc
@@ -412,7 +412,9 @@ std::array<double, 3U> MultiPartRain::extractDirection() const
   return result;
 }
 
-std::array<double, 3U> MultiPartRain::extractDirection(double sign_x, double sign_y, double sign_z) const
+std::array<double, 3U> MultiPartRain::extractDirection(double sign_x,
+                                                       double sign_y,
+                                                       double sign_z) const
 {
   double px, py, pz;
 
@@ -452,7 +454,6 @@ std::array<double, 3U> MultiPartRain::extractDirection(double sign_x, double sig
 
   return {px, py, pz};
 }
-
 
 TVector3 MultiPartRain::GenMomentum(const PartGenParam& param,
                                     const double& mass,
@@ -505,7 +506,6 @@ TVector3 MultiPartRain::GenMomentum(const PartGenParam& param,
     px = p[0];
     py = p[1];
     pz = p[2];
-
   }
 
   if (_debug) {
--- a/larsim/EventGenerator/MultiPart/MultiPartRain_module.cc
+++ b/larsim/EventGenerator/MultiPart/MultiPartRain_module.cc
@@ -72,6 +72,7 @@ public:
   TVector3 GenPosition();
 
   std::array<double, 3U> extractDirection() const;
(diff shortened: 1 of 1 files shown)


knoepfel:
@nvpm-lab, thank you for the PR.  The changes look sensible.  But do these changes affect the physics results?  If so, they should be presented at a LArSoft coordination meeting before being included as part of a LArSoft release.

nvpm-lab:
> @nvpm-lab, thank you for the PR. The changes look sensible. But do these changes affect the physics results? If so, they should be presented at a LArSoft coordination meeting before being included as part of a LArSoft release.

It doesnt affect the physics results in any ways, It just helps the generator run faster.

knoepfel:
Okay, please run the formatting code changes mentioned here: https://github.com/LArSoft/larsim/pull/164#issuecomment-3621118251

nvpm-lab:
> Okay, please run the formatting code changes mentioned here: [#164 (comment)](https://github.com/LArSoft/larsim/pull/164#issuecomment-3621118251)

Done!

lgarren:
trigger build

tomjunk:
Please check your distributions with plots!  I see a difference between the extractDirection() methods with and without arguments, when the cosmic distribution flag is set.  In one, the method without arguments, py = costheta, while in the method with arguments, pz = costheta.  In DUNE detectors, y is vertical, so it looks like the original method, without arguments, is the right one.  By the way, this is quite approximate and doesn't make sense underground.

larsim/EventGenerator/MultiPart/MultiPartRain_module.cc
```diff
@@ -411,6 +412,48 @@ std::array<double, 3U> MultiPartRain::extractDirection() const
...
+  if (_cosmic_distribution) {
+    double phi = fFlatRandom->fire(0, 2 * util::pi());
+    // Zenith angle theta in [pi/2, pi]
+    double costheta = -1. * fCosmicAngleRandom->fire();
+    double sintheta = sqrt(1 - costheta * costheta);
+
+    // To constrain the Original cosmic sampling to octant, multiply components by signs and use absolute values
+    px = cos(phi) * sintheta;
+    py = sin(phi) * sintheta;
+    pz = costheta;
```
tomjunk, line 430: It is py = costheta in MultiPartRain:extractDiretion() without arguments.  Physics has changed.

tomjunk (changes requested):
Please check py vs pz being set to costheta when _cosmic_distribution is true.

tomjunk:
I am also wondering what the source of your infinite loop was.  I think there might be a mistake in your use of clhep::RandGeneral in the original extractDirection() method.  RandGeneral returns a random number in (0,1):

https://mu2e.fnal.gov/public/hep/computing_retired/Random.shtml

whereas costheta is set thus:

double costheta = -1. * fCosmicAngleRandom->fire();

You may want never to explore upward-going cosmic rays, but depending on the values of x, y, and z, you could have a sign_z that makes your while loop never end.

nvpm-lab:
@tomjunk Ok i have managed to work around this! Thanks for the suggestions.

PR #165, opened 2025-12-25
Reimplementation of sim::BackTracker::HitToSimIDEs_Ps()

PetrilloAtWork:
This is an optional reimplementation of `sim::BackTracker::HitToSimIDEs_Ps()` taking advantage of the new `sim::SimChannel` interface from LArSoft/lardataobj#56.

It also changes the behaviour on invalid parameters from an implicit `std::terminate()` (`throw;`) to an actual exception, just for good practice.

On the reimplementation: much of the excised code stemmed from not assuming the elements of `sim::SimChannel` to be ordered in time, but this is an actual invariant of that class.
The rest of the overhead was working around lookup on the sorted data because of lack of an appropriate comparison, which is what LArSoft/lardataobj#56 provides.

### DIFF
--- a/larsim/MCCheater/BackTracker.cc
+++ b/larsim/MCCheater/BackTracker.cc
@@ -320,7 +320,10 @@ namespace cheat {
     if (start_tdc < 0) start_tdc = 0;
     if (end_tdc < 0) end_tdc = 0;
 
-    if (start_tdc > end_tdc) { throw; }
+    if (start_tdc > end_tdc) {
+      throw cet::exception("BackTracker") << "HitToSimIDEs_Ps(): Requested empty tick interval "
+                                          << start_tdc << " - " << end_tdc << "\n";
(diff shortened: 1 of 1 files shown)


knoepfel:
trigger build with pull requests #165, https://github.com/LArSoft/lardataobj/pull/56

tomjunk (approved):
Looks cleaner -- using the corresponding PR in lardataobj

tomjunk:
Regarding the DUNE test failure -- there are two failures -- a lingering one from PDVD, and a relatively  new one in DUNEFD simulation.  I ran the DUNEFD sim one interactively with a fresh build and it passed.  Hard to imagine how this PR would affect it however.  There was a change to the WireCell configs by Tracy but develop seems okay.  Not relevant for this PR.

lgarren:
approve

PR #167, opened 2026-04-24
Recovering charge deposited in CRP gap

dahendric:
In this new version the EField of the nearest active volume is used in order to properly calculate the amount of electrons and photons at the ionization and scintillation step


larsim/IonizationScintillation/ISCalcCorrelated.cxx
```diff
@@ -281,6 +300,48 @@ namespace larg4 {
     return angle;
   }
 
+  geo::TPCID ISCalcCorrelated::FindTPCForPosition(geo::Point_t const& p) const
+  {
+    constexpr double eps = 4.25; //max gap is 42mm
```
lpaulucc: Do we want this to be a hard coded number? Is there a change this can be changed in the future?

lpaulucc (commented):
Changes look ok and follow what has been discussed within the DUNE FD sim/reco group. Apart from the hard coded maximum gap value, I have no other comment.

larsim/IonizationScintillation/ISCalcCorrelated.cxx
tomjunk: This appears to search 4.25 cm in each direction from the mispoint of an energy deposition.  This can include volumes that are not active.  That said, VD detectors have "ghost" hits in the argon above the CRP that drift slowly with positive drift times.  There is also argon outside of the field cage, and in some places the electric field can get quite large outside the field cage.  Since VD has photon detectors on the field cage, is this a problem?

tomjunk: Actually, the PDS are on the cryostat membrane, but same question applies.

lpaulucc: Yes, the actual value of the electric field will impact the scintillation light production. Right now we are putting zero field everywhere outside the FC. It would be better to have a map of the EF from the HV WG but we have not been provided with one. From Daria's comments in the code, I understand that she looks for the nearest point in a TPC volume, so it should not include the region outside the FC. Is that correct @dahendric?

dahendric: Yes, the assumption is the following: if the point is not inside a TPC but there is a TPC on both left and right -then it is a gap. Moreover, this only happens if the SimEnergyDeposit was inside the LArG4DetectorServicevolEnclosureTPC (this is a paramter in the fhicl), which means that in principle the point should be inside the gap if it reached this part of the code

lpaulucc: Thank you, Daria. Could you put a comment on the unit to be used for the new fhicl parameter? Also, why do you multiply the number by 1.01?

dahendric: It was done to avoid any issues exactly at the border between the gap and active volume

dahendric: Added the comment. Also there are default values in dunesim/GapChargeTransport/GapChargeTransport_tool.cc, including the value of this fhicl parameter

knoepfel:
@dahendric, we've seen some conversation traffic here.  When do you anticipate moving this PR into "Ready" state?

dahendric:
> @dahendric, we've seen some conversation traffic here. When do you anticipate moving this PR into "Ready" state?

Hopefully this week. I don't see any unanswered questions left, and if the reviewers confirm that the current version is ok for them, I can remover the "draft" status

lpaulucc (approved):
Thank you, Daria!

tomjunk:
It's okay with me.  I'm not sure I understand what "left" and "right" mean in the comment above -- perhaps that there are TPCs on either side of the non-TPC point.

dahendric:
> It's okay with me. I'm not sure I understand what "left" and "right" mean in the comment above -- perhaps that there are TPCs on either side of the non-TPC point.

Yes, that is exactly what I meant, sorry for phrasing that in a confusing manner

knoepfel:
@dahendric, can we mark this as ready for review?

larsim/IonizationScintillation/ISCalcCorrelated.cxx
```diff
@@ -281,6 +301,48 @@ namespace larg4 {
     return angle;
   }
 
+  geo::TPCID ISCalcCorrelated::FindTPCForPosition(geo::Point_t const& p) const
+  {
+    double eps =
+      fMaxGap * 1.01; // in cm, default value is 4.2cm which is the largest gap in DUNE FD-VD
+    art::ServiceHandle<geo::Geometry const> geom;
```
knoepfel: I suggest making `fGeometry` a data member of `ISCalcCorrelated`, then you can repeatedly access it without having to recreate an `art::ServiceHandle` for every call to `FindTPCForPosition`.  The other uses of `art::ServiceHandle<geo::Geometry>` can then be adjusted to use only the one `art::ServiceHandle<geo::Geometry>` that is created whenever `ISCalcCorrelated` is constructed.

larsim/IonizationScintillation/ISCalcCorrelated.cxx
```diff
@@ -168,13 +170,22 @@ namespace larg4 {
...
+      // If we want to use the true electric field in the CRP gap, we need to check if the position is in the gap, and if so, use the field from the nearest TPC instead of returning 0.
+      if (!bool(tpcid)) {
+        tpcid = FindTPCForPosition(edep.MidPoint());
+        if (!tpcid) return 0.;
+      }
+    }
+    else {
+      if (!bool(tpcid)) return 0.;
+    }
     const geo::TPCGeo& tpcGeo = fGeometry->TPC(tpcid);
```
knoepfel: This pattern is repeated down below.  I suggest refactoring it into a dedicated function that can be used in two places (e.g.):

```cpp
geo::TPCGeo const& ISCalcCorrelated::TPCIDAtMidPoint(Point_t const& point)
{
  geo::TPCID tpcid = fGeometry->PositionToTPCID(edep.MidPoint());
  if (fUseGapAwareField and !tpcid) {
    // If we want to use the true electric field in the CRP gap, we need to check if the
    // position is in the gap, and if so, use the field from the nearest TPC instead of
    // returning an invalid one.
    return FindTPCForPosition(edep.MidPoint());
  }
  return tpcid;
}
```

and then it can be used like:

```cpp
auto tpcid = TPCIDAtMidPoint(edep.MidPoint());
if (!tpcid) {
  return 0;
}

geo::TPCGeo const& tpcGeo = fGeometry->TPC(tpcid);
// ...
```

This should simplify the repeated logic.

larsim/IonizationScintillation/ISCalcCorrelated.cxx
```diff
@@ -224,14 +235,23 @@ namespace larg4 {
...
+    if (fUseGapAwareField) {
+      // If we want to use the true electric field in the CRP gap, we need to check if the position is in the gap, and if so, use the field from the nearest TPC instead of returning 0.
+      if (!bool(tpcid)) {
+        tpcid = FindTPCForPosition(edep.MidPoint());
+        if (!tpcid) return 0.;
+      }
+    }
+    else {
+      if (!bool(tpcid)) return 0.;
+    }
```
knoepfel: See comment above about using a shared function (e.g. `TPCIDAtMidPoint`).

larsim/IonizationScintillation/ISCalcCorrelated.cxx
```diff
@@ -281,6 +301,48 @@ namespace larg4 {
     return angle;
   }
 
+  geo::TPCID ISCalcCorrelated::FindTPCForPosition(geo::Point_t const& p) const
+  {
+    double eps =
```
knoepfel: Suggest making this `const`.

larsim/IonizationScintillation/IonAndScint_module.cc
```diff
@@ -231,8 +235,19 @@ namespace larg4 {
...
 
       for (sim::SimEnergyDeposit const& edepi : *edeps) {
+        bool isGap = fGapTool && edeps.provenance()->productInstanceName() == fGapTool->Volume();
+
+        ISCalcCorrelated* fISAlg_gap = nullptr;
+
+        if (isGap) {
+          fISAlg_gap = dynamic_cast<ISCalcCorrelated*>(fISAlg.get());
+          if (fISAlg_gap) fISAlg_gap->SetGapAware(true);
+          if (fISAlg_gap) fISAlg_gap->SetMaxGap(fGapTool->MaxGap());
```
knoepfel, line 245: If `fGapTool->MaxGap()` is known early in the program and doesn't change, then I suggest updating `ISCalcCorrelated`'s constructor to take `MaxGap()` as another argument.  You could then remove the `SetMaxGap(...)` function.

larsim/IonizationScintillation/IonAndScint_module.cc
knoepfel: Does anything use the `ion_num_tmp` variable any more?

larsim/IonizationScintillation/IonAndScint_module.cc
```diff
@@ -284,6 +287,63 @@ namespace larg4 {
...
+                                  startTime_tmp,
+                                  endTime_tmp,
+                                  trackID_tmp,
+                                  pdgCode_tmp,
+                                  origTrackID_tmp);
+            continue;
+          }
+          else {
+            //GAP (light only)
+            simedep->emplace_back(ph_num, // all photons stay
```
knoepfel, line 311: Is this meant to be called unconditionally even if `moved > 0` is true?

larsim/IonizationScintillation/IonAndScint_module.cc
```diff
@@ -231,8 +235,19 @@ namespace larg4 {
...
+
+        if (isGap) {
+          fISAlg_gap = dynamic_cast<ISCalcCorrelated*>(fISAlg.get());
+          if (fISAlg_gap) fISAlg_gap->SetGapAware(true);
+          if (fISAlg_gap) fISAlg_gap->SetMaxGap(fGapTool->MaxGap());
+        }
+
         auto const isCalcData = fISAlg->CalcIonAndScint(detProp, edepi);
 
+        if (fISAlg_gap) fISAlg_gap->SetGapAware(false);
```
knoepfel, line 250: Ideally, `SetGapAware` would not be necessary...but I understand that if you do not know until line 238, then you may have to flip the state of `ISCalcCorrelated`.  There are a couple consequences to this:

1. This module cannot be executed on multiple events at the same time (e.g., `SetGapAware(...)` cannot be both `true` and `false` at the same time, which might be required if executing more than one event at the same time.
2. If `CalcIonAndScint` throws an exception, then `SetGapAware(false)` is never executed and the state of `ISCalcCorrelated` is maybe not set back to what it should be.

Item 1 is difficult to solve without rearranging how data products are retrieved.  Item 2 can be addressed by catching the exception and calling `SetGapAware` before the exception propagates further up the stack.  This might be more work than is practical, though.

Nothing obvious to do here regarding these two points...just something to be aware of.

knoepfel (commented):
@dahendric, thank you for the PR.  A few comments for your consideration below.  I'm happy to follow up on anything.

larsim/IonizationScintillation/IonAndScint_module.cc
dahendric: Yes. The idea is that the light should stay in the gap and be saved even if the charge is not moved due to being deposited too far from the active volume

larsim/IonizationScintillation/IonAndScint_module.cc
dahendric: Is it worth it to modify the constructor? I was trying to avoid it so that without the tool everything remains untouched. If I add it as an argument there should also be a default value, which seems a bit messy for the majority of cases where people don't use the charge recovery tool. However, that's up to you, I can change the constructor if needed

dahendric:
@knoepfel thanks a lot for reviewing this. I've updated the code following your comments except for the MaxGap question (see details in the comment above)

knoepfel:
trigger build

lgarren:
approve

PR #168, opened 2026-05-05
Update mpv to include the pdune beams

nvpm-lab:
Adds a configurable beam mode to the MultiPartVertex generator where particles are placed on the surface of an upstream hemisphere rather than uniformly within the TPC volume specifically designed for PDVD, PDSP, and PDHD.
**What changed**
BeamMode: when true, start positions are sampled uniformly on a hemisphere of radius BeamRadius centred on BeamEntrance, with the flat face flush with the detector entrance plane
BeamInwardDirection: unit vector pointing into the detector, defines hemisphere orientation (default [0,0,1] for PDVD)
BeamTargetRadius: particles are aimed at a random point within a disk of this radius around BeamEntrance, simulating a finite beam spot. Set to 0 to aim at the exact entrance point

New fcl parameters
```
BeamMode:            true
BeamEntrance:        [94.8, 142.6, 0.7]  
BeamRadius:          15.0                
BeamTargetRadius:    0.0                
BeamInwardDirection: [0.0, 0.0, 1.0]  
```

Adding beam mode generator display (BeamMode: true) and standard isotropic display (BeamMode: false) for comparison --> 20 events
image

### DIFF
--- a/larsim/EventGenerator/MultiPart/MultiPartVertex_module.cc
+++ b/larsim/EventGenerator/MultiPart/MultiPartVertex_module.cc
@@ -5,6 +5,17 @@
 //
 // Generated at Tue Dec 13 15:48:59 2016 by Kazuhiro Terao using artmod
 // from cetpkgsupport v1_11_00.
+//
+// Modified to support beam-mode hemisphere start position:
+//   particles are generated on the surface of a hemisphere of configurable
+//   radius, centered on a beam entrance point, with the flat face of the
+//   hemisphere flush with the detector entrance.  The momentum direction
+//   is always aimed from the surface point back through the entrance and
+//   into the detector.
+//
+// Further modified: allow random target point around beam entrance
+//   (BeamTargetRadius) so that particles are aimed at a random point
+//   within a disk of that radius, simulating a finite beam spot.
 ////////////////////////////////////////////////////////////////////////
 
 #include "art/Framework/Core/EDProducer.h"
@@ -70,9 +81,17 @@ public:
   TVector3 GenPosition();
   TVector3 GenBoost();
 
-  std::array<double, 3U> extractDirection() const;
+  // Beam-mode generators
+  TVector3 GenBeamEntrancePosition();
+  TVector3 GenBeamTarget() const;
+  std::array<double, 3U> GenBeamDirection(const TVector3& start_pos,
+                                          const TVector3& target) const;
+
+  std::array<double, 3U> extractDirection() const; 
   TVector3 GenMomentum(const PartGenParam& param, const double& mass);
   TVector3 GenMomentum(const PartGenParam& param, const double& mass, bool& same_range);
+  TVector3 GenMomentum(const PartGenParam& param, const double& mass,
+                       bool& same_range, const std::array<double, 3>& dir);
 
   double GenMomentumSF(const double& sf, const double& mass, const double& p);
   std::vector<size_t> GenParticles() const;
@@ -112,6 +131,25 @@ private:
 
   bool _use_boost;
   bool _revert;
+
+  // ---------------------------------------------------------------
+  // Beam-mode members
+  // ---------------------------------------------------------------
+  bool _beam_mode;
+
+  // Centre of the hemisphere = the nominal beam entrance point on the
+  // detector face, in LArSoft cm coordinates.
+  std::array<double, 3> _beam_entrance;
+
+  // Radius of the hemisphere [cm].  Particles are placed on this surface.
+  double _beam_radius;
+
+  // Unit vector pointing *into* the detector from the entrance face.
+  std::array<double, 3> _beam_inward_dir;
+
+  // Radius (in cm) of the disk in which the random target point is generated
+  // around _beam_entrance.  If zero, target = _beam_entrance exactly.
+  double _beam_target_radius;
 };
 
 void MultiPartVertex::abort(const std::string msg) const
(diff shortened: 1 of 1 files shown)


lgarren:
trigger build

larsim/EventGenerator/MultiPart/MultiPartVertex_module.cc
```diff
@@ -197,9 +273,9 @@ MultiPartVertex::MultiPartVertex(fhicl::ParameterSet const& p)
   _multi_min = std::max(_multi_min, multi_min);
   if (_multi_max < _multi_min) this->abort("Overall MultiMax <= overall MultiMin!");
 
-  if (!xrange.empty() && xrange.size() > 2) this->abort("Incompatible legnth @ X vector!");
-  if (!yrange.empty() && yrange.size() > 2) this->abort("Incompatible legnth @ Y vector!");
-  if (!zrange.empty() && zrange.size() > 2) this->abort("Incompatible legnth @ Z vector!");
+  if (!xrange.empty() && xrange.size() > 2) this->abort("Incompatible length @ X vector!");
```
tomjunk, line 276: I'd prefer these messages to indicate that they come from MultiPartVertex_module.cc -- is this something the abort() method has in it?  I'll see what abort() does.

tomjunk: Here's what abort() does:

https://github.com/nvpm-lab/larsim/blob/86862417c4861d49c988bf8c718f11b1332580bd/larsim/EventGenerator/MultiPart/MultiPartVertex_module.cc#L156

it just prints out the message (which doesn't say what module threw the error) and throws an error.  I like the error throwing, but better messages would be good.

tomjunk:
> BeamInwardDirection: unit vector pointing into the detector, defines hemisphere orientation (default [0,0,1] for PDVD)

Isn't PDVD angled at 45 degrees wrt the beam?  And is there any dip angle to the beam in NP02?  There's about a 10 degree dip in NP04 of the beam with respect to vertical.  A totally horizontal beam in NP02 would be more difficult to reconstruct than it has to be.

tomjunk:
> BeamMode: when true, start positions are sampled uniformly on a hemisphere of radius BeamRadius centred on BeamEntrance, with the flat face flush with the detector entrance plane
BeamInwardDirection: unit vector pointing into the detector, defines hemisphere orientation (default [0,0,1] for PDVD)
BeamTargetRadius: particles are aimed at a random point within a disk of this radius around BeamEntrance, simulating a finite beam spot. Set to 0 to aim at the exact entrance point

And I'm not sure I understand what the physics process is here.  Seems like this is a recipe for generating particles with a broad range of directions.  If I look at the old code, it seems isotropic, which doesn't sound plausible for any distribution of particles made when a beam hits material (I would expect more forward-going ones).  This modification looks more forward-going, but where is the justification for the angular distribution?  Is it meant to match some data?

larsim/EventGenerator/MultiPart/MultiPartVertex_module.cc
```diff
@@ -409,18 +573,6 @@ TVector3 MultiPartVertex::GenBoost()
   return TVector3(bx, by, bz);
 }
 
-std::array<double, 3U> MultiPartVertex::extractDirection() const
-{
-  double ct = fFlatRandom->fire(-1.0, 1.0);
```
tomjunk, line 414: This old code seems to generate isotropic directions -- why?  Half of the directions are backwards.  At least in the new code, according to the comments, directions are forwards, but they still have a lot of transverse components.

nvpm-lab:
> > BeamMode: when true, start positions are sampled uniformly on a hemisphere of radius BeamRadius centred on BeamEntrance, with the flat face flush with the detector entrance plane
> > BeamInwardDirection: unit vector pointing into the detector, defines hemisphere orientation (default [0,0,1] for PDVD)
> > BeamTargetRadius: particles are aimed at a random point within a disk of this radius around BeamEntrance, simulating a finite beam spot. Set to 0 to aim at the exact entrance point
> 
> And I'm not sure I understand what the physics process is here. Seems like this is a recipe for generating particles with a broad range of directions. If I look at the old code, it seems isotropic, which doesn't sound plausible for any distribution of particles made when a beam hits material (I would expect more forward-going ones). This modification looks more forward-going, but where is the justification for the angular distribution? Is it meant to match some data?

The isotropic angular distribution was agreed upon after discussion with the SPINE group. While one could model the exact beam angular distribution, the suggestion was to keep an isotropic distribution for both PDVD and PDHD for now, around the beam entrance. The key improvement here is not the angular distribution but the start position; rather than generating vertices at arbitrary points inside the detector volume, particles are constrained to originate near the beam entrance region. This gives SPINE a more physical training set where beam activity is localised to where the beam actually enters the detector, without committing to a specific angular model.

tomjunk:
It might be good to add a comment or even a message printed out in the constructor to tell users that this generator is to be used only for training AI/ML tools due to its flat(ish) phase-space coverage.

nvpm-lab:
Ok I will add that.

nvpm-lab:
`************************************************************
  MultiPartVertex: INTENDED FOR AI/ML TRAINING USE ONLY
  This generator samples phase space with broad, approximately
  uniform distributions (flat in KE/momentum, isotropic in
  angle, uniform in vertex position) to maximise detector-
  response coverage for training and validating reconstruction
  models.  It does NOT apply physical cross-section or flux
  weighting and is NOT suitable for physics analyses requiring
  a realistic kinematic or rate prior.
************************************************************`
Now the generator print this while running

tomjunk:
Did you make your error messages more descriptive regarding what threw the error?  It is awkward for people debugging to grep through the entire code base for a message they do not know the origin of.  Especially when messages are constructed from constant strings.  This can be made as simple as updating the abort() method to say what module the error is coming from.

Nivedpm6:
> Did you make your error messages more descriptive regarding what threw the error? It is awkward for people debugging to grep through the entire code base for a message they do not know the origin of. Especially when messages are constructed from constant strings. This can be made as simple as updating the abort() method to say what module the error is coming from.

I don't understand.., that was not an error . That is the description you told me to add so that people will know  that this is an ML training specific generator..

knoepfel:
@tomjunk and @nvpm-lab, what is the status of this PR?

tomjunk:
Sorry to be slow!    The error message I was concerned about is this one:

Incompatible length @ X vector!

which goes to cerr and then a std::exception gets thrown.  I think that message is enough if the module that originated it gets its name printed out as well, but otherwise it can be very mysterious, as the message could have come from anywhere in a large codebase and a user would have to grep through a lot of code to find it to understand how to debug it.  If the framework prints out the offending module's name when catching errors, that's fine.

tomjunk:
After checking some examples, it looks like _art_ does in fact print the offending module's name when catching exceptions. so I'll go ahead and approve it.

tomjunk (approved):
Approved since the framework provides more context when exceptions are thrown to help users debug the situation.  Some error messages look like they don't provide a lot of info.

lgarren:
code-checks

lgarren:
trigger build

lgarren:
approve

PR #173, opened 2026-07-27
Fix in-plane coordinate for orientation 2 in Rectangle_SolidAngle

sofiaksodhi:
For OpDetOrientation == 2 the arapuca plane is normal to z, so the in-plane coordinate paired with the window half-width o.w is y. The function used std::abs(v.Z()) as both the perpendicular depth and the in-plane width, so the returned solid angle had no dependence on the true in-plane position. Assign the in-plane coordinate d3 per-orientation alongside d1 and d2. No change to orientation 0 or 1, where z is in-plane.

### DIFF
--- a/larsim/PhotonPropagation/SemiAnalyticalModel.cxx
+++ b/larsim/PhotonPropagation/SemiAnalyticalModel.cxx
@@ -662,29 +662,32 @@ namespace phot {
     // to use
     double d1;
     double d2;
+    double d3; // in-plane coordinate paired with o.w
     if (OpDetOrientation == 2) {
       // lateral PD, arapuca plane fixed in z direction
       d1 = std::abs(v.X());
       d2 = std::abs(v.Z());
+      d3 = std::abs(v.Y());
     }
     else if (OpDetOrientation == 1) {
       // lateral PD, arapuca plane fixed in y direction
       d1 = std::abs(v.X());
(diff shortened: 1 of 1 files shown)


sofiaksodhi:
@pgreen135 Hi! I'm having some trouble adding you as a reviewer but here is my PR for the semi-analytical model as discussed this morning.

pgreen135:
The change looks good. Context: https://indico.fnal.gov/event/75260/contributions/346046/attachments/200180/279131/Photon%20detector%20orientation%20and%20solid%20angle%20in%20SemiAnalyticalModel%20for%20FD-VD%20%283%29.pdf 

To ensure this different orientation is being used, can you double-check that https://github.com/LArSoft/larsim/blob/5547c3c14b46cede6b91a703619c8ec5b3a819b8/larsim/PhotonPropagation/SemiAnalyticalModel.cxx#L781 correctly identify the short membrane arapucas?

sofiaksodhi:
Hello, yes I did check the short membrane arapucas get classified correctly (for v5 geometry). A simple count of how many channels are of what orientation yielded correct numbers: {0: 112, 1: 56, 2: 16}.

knoepfel:
code-checks

lgarren:
trigger build

lgarren:
trigger build

lgarren:
approve

